PR #2, opened 2016-09-03
Allow min and max dimensions for images to be swapped for portrait images

LostKobrakai:
Min/max dimensions on image fields are often used to determine a kinda basesize for image. This can be a pain for portrait images, which might be big enough, but are just a bit to narrow to fit for example the minimum width setting. 

This PR does allow those dimensions to be swapped for portrait images, so it's easier to make them work as well. The optimal solution could also be a min/max area size (e.g. 600x400 = 240000), but I'm not sure if that's a solution people would necessarily understand (rather a 3rd party module type of change).

### DIFF
--- a/wire/modules/Inputfield/InputfieldImage/InputfieldImage.module
+++ b/wire/modules/Inputfield/InputfieldImage/InputfieldImage.module
@@ -14,6 +14,7 @@
  * @property bool|int $maxReject Reject images that exceed max allowed size? (default=false)
  * @property int|string $minWidth Min width for uploaded images, smaller will be refused (default='')
  * @property int|string $minHeight Min height for uploaded images, smaller will be refused (default='')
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
Thanks, this seems useful, though thinking maybe it belongs as an Advanced tab setting, as maybe some would be confused by it. It's a bit cryptic to think about, but when the need is present, it seems very useful. The simpler we can make it I think the more would use it. What would you think about if we just collapsed it to a single setting of "Allow min and max dimensions for images to be swapped for portrait orientation" ?

LostKobrakai:
Yeah, that's probably not a bad idea. I'll take care of the changes sometime in the next few days.

LostKobrakai:
I've just updated the PR to only add a single checkbox. I'm not sure about adding it to "advanced" because there are only unrelated things there and I'm not sure people would expect it there.

ryancramerdesign:
Thanks @LostKobrakai I have merged this.

PR #9, opened 2016-09-23
fix missing prefix data- for minlength

jdiderik:
I found a type in the InputfieldPassword.js, causing the checks for password complexity fail.


jdiderik:
This seems to have been fixed in an update outside of this pull request.
Closing pull-request.

PR #12, opened 2016-10-01
Attachment support for core WireMail

LostKobrakai:
Implement attachment support for core WireMail so we can start using a common interface between different WireMail implementations.

### DIFF
--- a/wire/core/WireMail.php
+++ b/wire/core/WireMail.php
@@ -53,6 +53,7 @@ class WireMail extends WireData implements WireMailInterface {
 		'bodyHTML' => '',
 		'header' => array(),
 		'param' => array(), 
+		'attachments' => array(), 
 		);
 
 	public function __construct() {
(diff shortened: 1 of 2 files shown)


LostKobrakai:
I've not much experience in testing email stuff. It worked for me locally and the contentType/file_chunking part is from a StackOverflow topic, where I also didn't notice anyone talking about it not working. If you can point me to any specific things I could test I'd try to do that.

ryancramerdesign:
Tested out here and seems to work great. I have merged this to dev. Thanks. Looks like GitHub recognized this one, I think because you originated it on a "dev" branch as well.

LostKobrakai:
Yeah, I think so, too.

wire/core/WireMail.php
```diff
@@ -328,11 +328,12 @@ public function param($value) {
...
 	 *
 	 */
-	public function attachment($value) {
+	public function attachment($value, $filename = '') {
 		if(is_null($value)) {
 			$this->mail['attachments'] = array();
 		} else if(is_file($value)) { 
-			$this->mail['attachments'][] = $value; 
+			$filename = $filename ?: basename($value);
+			$this->mail['attachments'][$filename] = $value;
```
horst-n: Wouldn't it be better to swap $filename and $value, to avoid override a previously added file?
->attachment($value1, ''imag.jpg")
->attachment($value2, ''imag.jpg")

LostKobrakai: Do you see a use-case for multiple attachments named the same? I'd imagine 1 or 2 files max will be the bulk of mails. I see it rather than a case of being able to switch out a file without clearing the whole array of attachments.

horst-n: No I don't see this as a use case, but that it could be done by accident. Just asked, if swapping would preserve those accidents.

LostKobrakai: True. I've no hard feelings about that, so if @ryancramerdesign thinks switching would be the better call I'm fine with it.

PR #13, opened 2016-10-02
Allow modules to add exclusions to FileCompiler

mikerockett:
This change allows you to add `fileCompilerExclusions` to a module's info array. This is useful for those who have `vendor` and `lib` directories in their modules that may break after being compiled.

Refer: https://github.com/processwire/processwire-requests/issues/18

### DIFF
--- a/wire/core/Modules.php
+++ b/wire/core/Modules.php
@@ -4620,6 +4620,12 @@ class Modules extends WireArray {
 		
 		// if not given a file, track it down
 		if(empty($file)) $file = $this->getModuleFile($moduleName);
+
+		// don't compile when module compilation is disabled
+		if(!$this->wire('config')->moduleCompile) return $file;
+	
+		// don't compile core modules
+		if(strpos($file, $this->coreModulesDir) !== false) return $file;
 		
 		// check if a module doesn't want something (directory/file) compiled
 		$moduleInfo = $this->wire('modules')->getModuleInfoVerbose($moduleName);
@@ -4634,12 +4640,6 @@ class Modules extends WireArray {
 				}
 			}
 		}
-
-		// don't compile when module compilation is disabled
-		if(!$this->wire('config')->moduleCompile) return $file;
-	
-		// don't compile core modules
-		if(strpos($file, $this->coreModulesDir) !== false) return $file;
 	
 		// if namespace not provided, get it
 		if(is_null($namespace)) {
--- a/wire/core/Modules.php
+++ b/wire/core/Modules.php
@@ -4620,6 +4620,20 @@ class Modules extends WireArray {
 		
 		// if not given a file, track it down
 		if(empty($file)) $file = $this->getModuleFile($moduleName);
+		
+		// check if a module doesn't want something (directory/file) compiled
+		$moduleInfo = $this->wire('modules')->getModuleInfoVerbose($moduleName);
+		$exclusionsKey = 'fileCompilerExclusions';
+		if(isset($moduleInfo[$exclusionsKey]) && is_array($moduleInfo[$exclusionsKey])) {
+			foreach($moduleInfo[$exclusionsKey] as $exclusion) {
+				$exclusion = pathinfo($file, PATHINFO_DIRNAME) . '/' . $exclusion;
+				if(!in_array($exclusion, $this->wire('config')->fileCompilerOptions['exclusions'])) {
+					$this->wire('config')->fileCompilerOptions('exclusions', array_merge(
+						$this->wire('config')->fileCompilerOptions['exclusions'], [$exclusion]
+					));
+				}
+			}
+		}
 
 		// don't compile when module compilation is disabled
 		if(!$this->wire('config')->moduleCompile) return $file;


mikerockett:
@ryancramerdesign - when you have a look at this PR, please would you also check out [this forum thread](https://processwire.com/talk/topic/14408-exhausted-memory/)?

The code in the PR was working, but something changed and now `$moduleName` recursively returns the name of a single module (via `var_dump`), either causing a memory max-out or a timeout.

In the meantime, I'm including my FileCompiler exclusion in the site config again so I can continue developing.

mikerockett:
**Update:** I've somehow managed to get this to work, but will need to test on a new install again (beginning to end) to ensure the smooth flow of things. When I test, I'm going to commit my PR code to the setup before installing.

ryancramerdesign:
Thanks @mikerockett – what do you think about the existing exclusion method of adding this comment anywhere in a file?
~~~~
// FileCompiler=0 
~~~~
This is the method I've been using anyway. This prevents compilation of the file, as well as any other files include()'d or require()'d from the file, basically removing the FileCompiler from the picture in one simple step. If you were dealing with lots of files you wanted to prevent from being compiled, then you could move those require()/include() statements to a separate file that had the comment, and then just include that one file that has the comment.

mikerockett:
Hi Ryan,

I'd actually completely forgotten about that comment, which is handy in general situations. However, it doesn't become useful in the situation where `vendor` directory is used and those files could be updated at any time. This is why I suggested that the exclusion be added at compile time, allowing the dev to specify a file, group of files, or a directory in a single location.

ryancramerdesign:
Good point. Actually I'm wondering if /vendor/ dirs off modules should just
be a global exclusion, as it seems like that's something we'd always want
to exclude?

mikerockett:
Not a bad idea, and I did think about it. But then I thought about PW being unopinionated and allowing devs to choose. However, `vendor` is standard and pretty much anything devs include in `lib` and `classes` directories could also be pulled in by Composer.

An additional benefit of making `vendor` a global exclusion is speed. In such a case, the system need not check module config, which uses additional resources.

So yeah, thats a 👍 from me.

teppokoivula:
This is (obviously) just an opinion, but personally I would vote against globally excluding directories based on a name alone. Out of the directories mentioned here only `vendor` is something I haven't yet used in my own projects, but even there I'd argue that ...

a) this sort of thing happening automatically can be confusing and difficult to debug,
b) there's an off chance of name collision with something completely unrelated, and
c) it might not even solve all issues (as pointed out by @mikerockett above).

In my opinion a configurable list of exclusions would be safer, more obvious, and more flexible :)

mikerockett:
> a) this sort of thing happening automatically can be confusing and difficult to debug,

Yes, I agree with this - not everyone will know that vendor is automatically excluded.

> b) there's an off chance of name collision with something completely unrelated, and

Do you mind elaborating on this? Perhaps I'm missing your point...

---

For me, I do actually think configuring this is the better route to take, giving consideration to the developer and their elaborate choices (these are expected in the development process). The only issue could perhaps be related to performance. For example, does this method get called for every request? Surely this is done before the compiler checks for already-compiled files? If this is the case, loading up config data for every module on every request would be overkill. The environment in which I'm developing Jumplinks has no other modules installed (except for Tracy), so I haven't tested possible impacts.

teppokoivula:
> Do you mind elaborating on this? Perhaps I'm missing your point...

The point was just that if I were to include a directory called `vendor`, files in it would be skipped, whether or not it has anything to do with Composer. I don't think I've ever personally used such a folder name, but it's not like Composer has a monopoly on "vendor" :)

To be honest I haven't even checked what the implementation looks like, but I was under the impression that this wouldn't happen on every page load.

mikerockett:
> The point was just that if I were to include a directory called vendor, files in it would be skipped, whether or not it has anything to do with Composer. I don't think I've ever personally used such a folder name, but it's not like Composer has a monopoly on "vendor" :)

I see - that makes sense. Have never seen vendor used for other things. But it is a common term nonehtless.

Another thing I'd like to add: Whilst this is very, and I mean *very* rare (the way I see it, especially in the PW community), some folks do change the names of their Composer `vendor` dirs. I can't really imagine a good reason as to why one would do such a thing (yes, I am guilty of it - and realised there was no point). That aside, given that customisation *can* and *may* happen from that point-of-view, it adds a +1 on the side of configuring it as per my PR.

It would be awesome to see PW having a built-in `composer install` implementation so that modules would not need to be shipped with vendor dirs. All I'd need to do is ship my composer.json, and the dependencies would be loaded in the module dir, and then PW would automatically add `vendor` to `fileCompilerExclusions`. That would be magical.

harikt:
Actually I have a question : 

Are you saying about the `vendor` or `lib` folders inside a module or the root `vendor` folder ?

If it is 3rd party library carried inside a module I would argue to let composer install it on root of the project `vendor` so that PW don't need to worry about the compile step. That is why we added the composer installer for pw module. It also helps 3rd party libraries to get updated without the author of the module updating and releasing when there is security issues related to the 3rd party libraries.

mikerockett:
> If it is 3rd party library carried inside a module I would argue to let composer install it on root of the project vendor so that PW don't need to worry about the compile step.

It is indeed a good idea, but Composer does not allow you to run multiple versions of each package concurrently. The reason I bring this up is that I'm using older versions of certain modules to allow for PHP 5.4/5.5. If another module needs a newer version, then it should be able to use its own.

However, I don't see Composer packages being used very widely within modules, but we should try and make this future-proof.

harikt:
> Composer does not allow you to run multiple versions of each package concurrently.

I wonder how the modules can itself solve the problem in that case. Reason is PW any how will load `vendor/autoload.php` if it is there. And a call to a class will try to load the class. So how do you guarantee your class / object inside the module is called? Only thing I see widely used one which have trouble is guzzle. Apart from that I don't see or have came across the issue in my composer usage life :) .

> However, I don't see Composer packages being used very widely within modules, but we should try and make this future-proof.

Yes, it is still in the process of adopting in PW I guess :) .

LostKobrakai:
> It is indeed a good idea, but Composer does not allow you to run multiple versions of each package concurrently.

As I already posted on the forums this is not possible no matter how you turn it. If you install a library in multiple versions you'll get collisions no matter where you store the code. 

> However, I don't see Composer packages being used very widely within modules, …

I'd think that's because bundling dependencies never seemed like a good idea – at least to me. I'll certainly be using the option more with 3.0 having support for composer (already have about 3 in alpha/beta status).

>It would be awesome to see PW having a built-in composer install implementation so that modules would not need to be shipped with vendor dirs.

There's no way to reliably detect a globally installed composer package. So this would mean we either need to bundle the composer package into processwire (probably a bad idea), install it as part of the installation process or the user does need to install composer in the project manually. I feel like those might be the only ways to keep sanity for installing composer dependent modules through the admin backend.

> Only thing I see widely used one which have trouble is guzzle.

Yeah, same for me.

mikerockett:
> As I already posted on the forums this is not possible no matter how you turn it. If you run a library in multiple versions you'll get collisions no matter where you store the code.

Of course, right - train of thinking went the wrong way again. That said, surely it would still be useful in the case of modules that do not autoload on every request? That said, if one module autoloaded and requires a dependency that could possibly conflict, two non-autoload modules would face a potential problem.

> ... or the user does need to install composer in the project manually.

Do you mean the user/dev would have to install the module and then `composer install` inside the module dir? If so, it is the route I prefer so that the dependencies need not be bundled with the module itself. Altnernatively, the module would be installed via `composer require rockettpw/jumplinks`, which would install the module and bring in the dependencies. However, as above, there's still room for conflict, which I guess is inevitable in the PW setup.

Bolt CMS uses Composer for its extension handling. I'm not quite sure how it handles potential conflicts, but I'd assume that extension devs tend to use the latest and greatest of everything, and extensions therefore share the same dependency base. However, this could present a potential problem for some folks in terms of PHP compatibility.

As mentioned, I don't want JL2 to raise the minimum PHP version for those who are using hosts that simply don't offer newer versions. If I were to depend on masters of these dependencies, the min. PHP version required would be 5.6.4. This topic on its own could well branch off into a new one, and a _long_ discussion could be had about it (I don't see why hosts should not be up to date, but I wouldn't want to be unfair to users/devs that use these hosts).

Hehe, I guess it would have been neat if Composer could prefix everything in a vendor dir with a particular namespace, but that wouldn't actually be suitable as not all files use a namespace. It would solve a few problems, however.

harikt:
> Do you mean the user/dev would have to install the module and then composer install inside the module dir? 

No.

> Altnernatively, the module would be installed via composer require rockettpw/jumplinks, which would install the module and bring in the dependencies.

Yes. Ryan has already blogged about the same with some examples. ( Don't have the link for the current time) . I have also wrote about this some years ago : http://harikt.com/blog/2013/11/16/composer-support-for-processwire-modules/

Hope that helps.

mikerockett:
@harikt - Yes, and I do honestly prefer that approach for Composer-dependent modules. Haven't actually had the chance to thank you for putting that all together, it really is excellent. 👍

With that said, modules will still want to have per-file/directory exclusions in certain circumstances, I am sure. So this PR _could_ still stand- but, as these would more than likely not be package-manager-driven, `// FileCompiler = 0` is maybe be the better route to take, unless an entire directory needs to be excluded from compilation.

In terms of JL2, I think I'm going to go the pw-module route. For me, it makes the module future-ready. Question is, what about the folks that are upgrading from JL1 to JL2? I take it the `composer require` method would be suitable? No way to prevent installation/upgrade from the Modules admin page until such time as Composer would have full support without the need for CLI access (like Bolt CMS)?

harikt:
I don't know what JL2 is. What is it actually used for ? 

Is it available in packagist ?

I also don't know how Bolt cms work. Probably we can take something from them if people like those feature :) .

mikerockett:
> I don't know what JL2 is. What is it actually used for ? 

Sorry - I thought you may have picked it up from the forums. It's referring to my module, Jumplinks - I'm busy building version 2. It isn't available as yet - still under development.

mikerockett:
> In terms of JL2, I think I'm going to go the pw-module route. For me, it makes the module future-ready. Question is, what about the folks that are upgrading from JL1 to JL2? I take it the composer require method would be suitable? No way to prevent installation/upgrade from the Modules admin page until such time as Composer would have full support without the need for CLI access (like Bolt CMS)?

@harikt/@ryancramerdesign: Do you perhaps have any feedback on the above?

harikt:
> Question is, what about the folks that are upgrading from JL1 to JL2? I take it the composer require method would be suitable? No way to prevent installation/upgrade from the Modules admin page until such time as Composer would have full support without the need for CLI access (like Bolt CMS)?

Sorry. I probably missed to understand. Do you mean boltcms have some way to install the composer dependencies via admin, and you are looking for processwire to do the same ? Then I am not sure when will such a feature will come in PW admin though.

mikerockett:
> Sorry. I probably missed to understand. Do you mean boltcms have some way to install the composer dependencies via admin, and you are looking for processwire to do the same ? Then I am not sure when will such a feature will come in PW admin though.

Yes, that's the idea. For the most part, I'm worried about the those upgrading from Jumplinks 1 to 2 via the admin panel if I were to make it a pw-module, installlable via `require`.

I have updated my roadmap regarding changes to the dependencies I require that brought up this PR and related issue. That said, without the suggested changes, the vendor directory would still be scanned, which could take a while, and potentially max out resources on some setups.

wanze:
👍  for the possibility to tell the compiler to exclude certain directories. My TemplateEngineFactory module has the problem that twig templates are getting compiled by Pw (they shouldn't) which leads to errors. Current solution is to add `{# FileCompiler=0 #}` to *all* twig templates. So I would need the possibility to exclude the directory containing the twig templates from being compiled.

Related issue:
https://github.com/wanze/TemplateEngineTwig/issues/3

wanze:
@ryancramerdesign Any news on this?

mikerockett:
@wanze - Ryan mentioned in another thread (can't remember which one) that he is working on it. Not sure if this PR will make it as is, if or Ryan has a better way of handling it. Personally, I like the flexibility of adding this to a module's config data.

ryancramerdesign:
I've added one option here, which enables you to disable the FileCompiler at the time you include other files. You just add a comment within the `include()` or `require()` (or `_once` variations) statement that has `/*NoCompile*/` either at the front or end of it: 
~~~~~
require(/*NoCompile*/ __DIR__ . '/file.php');
require(__DIR__ . '/file.php'/*NoCompile*/);
~~~~~
This is alternative to the existing `FileCompiler=0` that can be added within any php file. And PHP files that declare a namespace are already always excluded. So now we've got three options for exclusion. The question is, do we still need another? For instance, I'm not sure any of these solve what @wanze was looking for (though using a non-php file extension for Twig templates might be a good approach too). 

While I like what's in the PR here and it seems well thought out, I feel that it's better if we don't store this in the verbose module information for a few reasons. One is that the verbose module info is kept separate from the regular module info so that it doesn't have to load it all the time (because it consumes significant memory). If it has to load it to query for FileCompiler exclusions, then it basically has to load it all the time. Another is that it's cached, so the exclusions wouldn't be known until a Modules > Refresh is done, which could be headache when you are developing a module and need to change anything about whether a file/dir is compiled or not. I think it's best if the exclusions are kept as meta data in something that FileCompiler is already processing, so that it's picked up immediately. That's essentially what the `/*NoCompile*/` and `FileCompiler=0` are.

mikerockett:
@ryancramerdesign - That looks perfect - thank you very much. 👍 

And thanks for the insight on how it all works, specifically regarding the cache.

Will test this out shortly.

mikerockett:
Looks like this works - it copies the contents of the module's directory to the cache directory still, but this happens pretty quickly (I assume this copy-operation is expected even if the files within are not compiled). Also not getting the errors anymore.

Thanks, Ryan - this is great.

I think I will keep Jumplinks 2 compatible with 2.8.x and >= 3.0.43.

PR #14, opened 2016-10-03
added support for CLI

horst-n:
without that, PHP throws notices when try to access the $_SERVER['REMOTE_ADDR'], because it is not set in CLI-Environment.

### DIFF
--- a/wire/core/Session.php
+++ b/wire/core/Session.php
@@ -651,7 +651,11 @@ class Session extends Wire implements \IteratorAggregate {
 	 */
 	public function getIP($int = false, $useClient = false) {
 
-		if($useClient) { 
+		if('cli' == php_sapi_name()) {
+			// when accessing via CLI Interface, $_SERVER['REMOTE_ADDR'] isn't set and trying to get it, throws a php-notice
+			$ip = '127.0.0.1';
+			
+		} else if($useClient) { 
 			if(!empty($_SERVER['HTTP_CLIENT_IP'])) $ip = $_SERVER['HTTP_CLIENT_IP']; 
 				else if(!empty($_SERVER['HTTP_X_FORWARDED_FOR'])) $ip = $_SERVER['HTTP_X_FORWARDED_FOR'];
 				else if(!empty($_SERVER['REMOTE_ADDR'])) $ip = $_SERVER['REMOTE_ADDR']; 


horst-n:
I don't know about the other IP handling in PW. I only discovered it trying to read $_SERVER["REMOTE_ADDR"], when operating via CLI (accessing via bootstarp script from the (windows) console). And regardless how you have setup your OS system, when CLI access is detected here in Session, it always returns 127.0.0.1, what is the localhost loopback IP. There is no harm here, even if a IP-v4 address for localhost isn't set in your system. (what I hardly can believe)
My PR simply want to avoid a PHP Notice, and return back something recallable for this method. How and where converting / handling IPv4 / IPv6 is done: I don't know. But adding CLI-Access detection will not change anything, just suppress the PHP Notice.

matjazpotocnik:
Now I see that I "abused" your PR, please apologize. Just wanted to draw attention to the fact that eventually we will need IPv6 support. On my computer _SERVER["REMOTE_HOST"] = ::1, _SERVER["REMOTE_ADDR"] = ::1 and PW can't handle this (when needed).

horst-n:
Maybe something with php functions ``inet_pton`` or ``inet_ntop`` can help with IPv6 here:
if(defined('AF_INET6')) // use IPv6 functions
else // use IPv4 functions
read more here: http://php.net/manual/en/function.inet-pton.php

ryancramerdesign:
@matjazpotocnik Yep we do need to add IPv6. I've not yet looked into it enough to know what's necessary, but will plan to do so. 

@horst-n Thanks looks great – I will go ahead and merge this one. Since we using a CLA with this new repo, would you mind completing the one linked from our CONTRIBUTING.md file? https://github.com/processwire/processwire/blob/master/CONTRIBUTING.md#pull-requests-prs Thanks.

horst-n:
@ryancramerdesign Done. I have signed the CLA.

ryancramerdesign:
Thanks Horst. Just looking at all the possible return values for the php_sapi_name function, I'm wondering if we should go with a different if() condition before populating the 127.0.0.1, like `if(empty($_SERVER['REMOTE_ADDR']))` ?

horst-n:
As far as I know, the only constant return value is lowercase "cli" for CLI. But to be on the safe side, it may be better checking $_SERVER["REMOTE_ADDR"].
+1

ryancramerdesign:
Thanks Horst, this has been added to the dev branch.

teppokoivula:
Just for the record, there are some borderline case where REMOTE_ADDR is empty but the request didn't necessarily originate from local environment. I'm not sure if this really has to be accounted for, but (for an example) I've heard some reverse proxy setups omitting this :)

This won't be an issue unless someone really needs to see which requests originate from the local address, but if trusted blindly it may result in strange behaviour.

horst-n:
So best thing would be, first check if REMOTE_ADDR is empty and additionally check for "cli" == php_sapi_name() ?

PR #17, opened 2016-10-12
Support utf8 characters in WireMail

BitPoet:
Use quoted printable encoding for text and html body parts to allow non-ascii characters.
Also encode the subject and the name parts of from: and to: headers using quoted printable according to rfc2047.


ryancramerdesign:
Thanks–great addition. Before we merge this to master, can you complete the 2nd item for CLA in our pull-requests contributing document file here: https://github.com/processwire/processwire/blob/master/CONTRIBUTING.md#pull-requests-prs

Thanks.

PR #19, opened 2016-10-20
Subject wrapping, mbstring support, nested content parts

BitPoet:
Support wrapping of long subject headers, using mb_encode_mimeheader if
available, otherwise fall back to an custom encoding/wrapping function
that uses PHP builtins.
Nest plaintext+html bodies in their own multipart-alternative header if
attachments are present so that the display of the html part still has
precedence over the plaintext part.
Put quotes around name parts after quoted-printable encoding of the name
in name+address combinations.

### DIFF
--- a/wire/core/WireMail.php
+++ b/wire/core/WireMail.php
@@ -129,13 +129,14 @@ class WireMail extends WireData implements WireMailInterface {
 		$email = $this->sanitizeEmail($email); 
 		if(!strlen($name)) return $email;
 		$name = $this->sanitizeHeader($name); 
+		$delim = '';
 		if(strpos($name, ',') !== false) {
 			// name contains a comma, so quote the value
 			$name = str_replace('"', '', $name); // remove existing quotes
-			$name = '"' . $name . '"'; // surround w/quotes
+			$delim = '"';  // add quotes
 		}
 		// Encode the name part as quoted printable according to rfc2047
-		return $this->quotedPrintableString($name) . " <$email>";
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
Thanks @BitPoet, looks good, how much has this been tested thus far? Also had a question about this line:
~~~~~
if(extension_loaded("mbstring") && false == true) {
~~~~~
What's the purpose of the `false == true`? It seems like it would cause that block to always be skipped over. Or is that just an extra from testing?

BitPoet:
Sorry, yes, the `&& false == true` is a leftover and shouldn't be there at all. I've amended the pull request.

So far, I've ran a suite of tests with different combinations of subjects, bodies and attachments, sending each one through different servers and opening them with Outlook, Thunderbird and Gmail. It might perhaps make sense to test the patch on a Linux system too, as I currently only have Windows available. I don't foresee any problems there, but better save than sorry.

ryancramerdesign:
Thanks. Another question for you. On this line, the variable `$offsetHdr` is undefined. Should it instead be `$maxEffLen`? No need to update the PR if that's the case, I can easily change that here, but wanted to check.

ryancramerdesign:
Sorry I hit the comment button too soon, here's the line:
~~~~
// Force cutting of subject since there is no whitespace to break on
$part = substr($input, $maxlen - $offsetHdr);
~~~~~

BitPoet:
Yes, this should be
```
$part = substr($input, $maxEffLen);
```
I ran a diff between the tested and the committed version and they are identical now, so this should definitely be the last embarrassing slip-up I made with this commit.

PR #23, opened 2016-10-26
added * in line 170 so comment is parsed correctly

clsource:
Line 170 lacks an asterisk *
https://github.com/processwire/processwire/blob/master/wire/core/Functions.php#L170

This just add that in order to github not complaining
and comment could be parsed correctly


ryancramerdesign:
Thanks, I have added this

PR #28, opened 2016-11-05
Add support for optional overrides array to $input->queryString()

teppokoivula:
This feature is something I've found pretty useful when dealing with query strings: overriding existing GET params with custom ones, or adding new ones. While this can be done on a case-by-case basis with just a few lines of code, I thought it might also make sense as a core addition:

```
echo $input->queryString(); // action=edit&template=home

echo $input->queryString(array(
    'action' => 'rename',
    'limit' => 5,
)); // action=rename&template=home&limit=5

echo $input->queryString(array(
    'action' => null,
)); // template=home
```


ryancramerdesign:
Thanks Teppo, I have added this.

teppokoivula:
Thanks, Ryan!

PR #30, opened 2016-11-07
Set cookie domain on cookies

derixithy:
Allow the option to set the cookie domein so one session can be used over different sub-domains.


adrianbj:
@derixithy - please make PRs against the dev branch - thanks!

PR #31, opened 2016-11-07
Sanitizer::string small refactoring

pine3ree:
Since we are doing type-checking we are dealing with mutually excluding conditions.

So far the code blocks for non-string type matching will transform `$value` into a string anyway so if we have an object, the null value or a bool we would get a string and any further test would be not needed for it. W only need to test the unprocessed value, i.e. other conditions on the input value.

Now the code reads like this:
if input value is:
object? => get a string
or null? => get an (empty) string
or bool? => get a (numeric|empty) string
or array? => build a string
or anything else but a string => cast to string

Off-Topic:
shouldn't be better to use `null === $var` / `null !== $var` instead of calling `is_null`. Inside a function that can be called many times it can make a diifference in processing time since calling a function is always more expensive than a simple (strict) comparison, even though i agree to be more consistent with other type-checking.

kind regards

### DIFF
--- a/wire/core/Sanitizer.php
+++ b/wire/core/Sanitizer.php
@@ -1823,9 +1823,11 @@ class Sanitizer extends Wire {
 			$value = "";
 		} else if(is_bool($value)) {
 			$value = $value ? "1" : "";
+		} else if(is_array($value)) {
+			$value = "array-" . count($value);
+		} else if(!is_string($value)) {
+			$value = (string) $value;
 		}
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
Thanks, I have added this. I think originally there was some code in there where an object can be converted to an array, so the array check had to be outside the if() block. That's clearly not the case anymore, if it ever was. Regarding is_null() vs === null, it's a micro-optimization which is not worth consideration, unless we're specifically building an is_null() vs === null application. :)

PR #33, opened 2016-11-08
Remove duplicate settings of template tags. It is already set 7 lines above.

adrianbj:
I think this makes sense and I don't think it should matter which one is removed.


teppokoivula:
@adrianbj just checking: did you notice that there are two apparently unrelated commits in this PR? :)

adrianbj:
Thanks @teppokoivula - I branched from the wrong base :)

I'll close this and prepare a new one!

LostKobrakai:
You can always update the branch locally and force push to github. This will update the PR.

PR #34, opened 2016-11-08
Remove duplicate line for setting template tags.

adrianbj:
I don't think there is any reason for the one I removed and I can't see why it matters which one is removed.


ryancramerdesign:
Thanks Adrian, I have added this. Since it's just one line I didn't go through the whole pull, merge, etc. process and instead just deleted the line.

PR #35, opened 2016-11-09
turkish chars added to defaultReplacements array

borantula:
Special letters in Turkish alphabet (see below) added to default replacements array as they were becoming - in page names.

ı=i
ğ=g
İ=i
Ç=c
ş=s
Ş=s


ryancramerdesign:
Thanks @borantula it looks like none of these additions conflict with existing defaults, so we should be able to add this. Before I pull this in can you complete item #2 from our contributing file? https://github.com/processwire/processwire/blob/master/CONTRIBUTING.md#pull-requests-prs

borantula:
I read and agreed, with a hope to send more PR's :)

ryancramerdesign:
Thanks, this has been added. Also for future PRs use the dev branch, just in case (since that's what we merge to). In this case it didn't matter though.

PR #38, opened 2016-11-22
ProcessPageEditLink: edit link text feature

rolandtoth:
Adds a field to enter link text in the edit link modal. The new field is available only for non-image links.
Also, the link previewat the bottom shows the link text too.

See more at
https://processwire.com/talk/topic/14820-add-link-text-to-processpageeditlink/


ryancramerdesign:
Thanks @rolandtoth this seems like a good addition and I'll be looking closer and adding it hopefully this week. One question is about link text that has markup in it, like if you are linking `text <em>like</em> this`, have you given thought to how it handles the `<em>`?

rolandtoth:
Good catch - they are stripped currently. I'll look into it in the following days, I think it's easy to fix.

ryancramerdesign:
I kind of think that if the link text has markup in it, the feature should
just not activate. I don't think there's a good way to handle markup here,
because really the only way it could be handled would be to show them the
markup and let them edit it. That's okay for us, but probably a non-starter
for the end users. If it just simply disabled the link text feature if
there was markup, then it would just mean they have to edit it in CKEditor
instead, as one would do now.

rolandtoth:
Well the opposite might be more productive, letting to add a link even if the markup is lost on add. Then the user could re-add the markup if needed.

This is the way the original CKEditor works too:
http://ckeditor.com/features

Toutouwai:
Would be great to get this merged into the core. It's so easy to forget what text you've highlighted when selecting from a long list of files to link to, for instance.

ryancramerdesign:
Thanks @rolandtoth this has been added

PR #39, opened 2016-11-26
Fix typo and make tab navigation always visible

teppokoivula:
This commit fixes a typo that prevented using the module and also makes tab navigation visible even when active tab is empty. Currently tabs just disappear if you select a tab with no comments.

Note that empty `<form>` element around `<h2>` prevents minor visual glitch: tabs without wrapping form element would be 2px lower, which would look a bit strange. This feels like a hack, but I'm not familiar with the styles of default admin theme and didn't want to risk breaking anything else.

Fixes processwire/processwire-issues#95


ryancramerdesign:
Thanks Teppo! I have added this.

PR #37, opened 2016-11-20
Improve accessibility in site profiles

marcus-herrmann:
Hi Ryan,

this is an updated version of this PR from August: https://github.com/ryancramerdesign/ProcessWire/pull/1758 and we spoke about it here: https://github.com/processwire/processwire/pull/1.

Since I found some things to improve I created this PR here anew albeit initially taking your offer of cherry picking the old one (hopefully you did not already start :/ )

CLA is signed. I'm happy to answer any questions related to the improvements :)

Best,
Marcus

PR #40, opened 2016-11-27
Add defaults, labels, markers for accessibility to MarkupPagerNav

marcus-herrmann:
Hi Ryan,

I somehow believe that you will refactor this PR ;) 
Nevertheless - I hope you'll see where I'm aiming at: Improving MarkupPagerNav with non-visual markers and labels for assistive technologies, making some of them configurable.

Reference for all this was http://www.a11ymatters.com/pattern/pagination/

Like last time, I signed the CLA* and I'm happy to answer questions about this merge request.

Kind regards from Berlin,
Marcus

*) I just assumed this is necessary for every PR - but is it?
*) edit: I also updated PR #37 regarding aria-current (Background: www.w3.org/TR/wai-aria-1.1/#aria-current) and [this article, Section: aria-label, the problem](http://www.heydonworks.com/article/the-accessible-current-page-link-conundrum) explains why I chose to label pagination differently from site profile navigation.

### DIFF
--- a/site-beginner/templates/_func.php
+++ b/site-beginner/templates/_func.php
@@ -30,7 +30,7 @@ function renderNav(PageArray $items) {
 		// render markup for each navigation item as an <li>
 		if($item->id == wire('page')->id) {
 			// if current item is the same as the page being viewed, add a "current" class to it
-			echo "<li class='current'>";
+			echo "<li class='current' aria-current='true'>";
 		} else {
 			// otherwise just a regular list item
 			echo "<li>";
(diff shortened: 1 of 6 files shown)


ryancramerdesign:
Thanks @marcus-herrmann I have added this. I ended up changing a few things around a bit and combined it with some other MarkupPagerNav updates. Let me know if you find I've screwed anything up. Thanks. 

Regarding your question on the CLA, it's not necessary to complete it more than once, as it applies to any future contributions through here as well.

marcus-herrmann:
Works as intended :)

PR #41, opened 2016-12-11
Better size balance between portrait and landscape images

Toutouwai:
Also fixes invalid !important rules that were missing an exclamation mark.

### DIFF
--- a/wire/modules/Inputfield/InputfieldImage/InputfieldImage.scss
+++ b/wire/modules/Inputfield/InputfieldImage/InputfieldImage.scss
@@ -644,7 +644,7 @@ $itemPadding: 0.4em;
 			top: 0;
 			left: 0;
 			float: left;
-			background: initial;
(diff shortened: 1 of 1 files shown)


Toutouwai:
Before:
images-before

After:
images-after

I'm not a big fan of the checkerboard background (would prefer a solid light grey myself) but thought I'd better leave that alone.

anilex:
Without an intention to be in opposite side, I prefer the "before" layout, no matter if "After" is with or without solid gray background.

horst-n:
I'm against the checkerboard, as it doesn't work with transparent images, what also doesn't work with solid colors. But definetly there is a need for better communication of the dimensions / proportions of the images in the thumbnail views!
There are some examples in the forum already, with screens and explanations: https://processwire.com/talk/topic/13808-background-for-png-images/#comment-124313 
(also following and/or above posts)

PR #42, opened 2016-12-13
Remove reference to specific version in install.php

LostKobrakai:


ryancramerdesign:
Thanks this has been added.

PR #46, opened 2016-12-23
Fix for incorrect path returned from $input->url() when called before page is available.

adrianbj:
It currently returns: `//admin/page/edit/-id-1111/?id=1111`
Instead of the correct: `/admin/page/edit/?id=1111`

This appears to fix everything correctly.


adrianbj:
Great - thanks @ryancramerdesign

PR #48, opened 2016-12-29
Fix issue #141 search input hides on key up with AdminThemeReno

gmclelland:
This PR fixes https://github.com/processwire/processwire-issues/issues/141.


ryancramerdesign:
This PR has been added (albeit in a different way) via the referenced commit. Thanks.

PR #49, opened 2017-01-11
Make InputfieldSelector properly respect "allowSystemTemplates" setting.

adrianbj:


adrianbj:
AJAX calls (eg after selecting "Template" from first dropdown) were calling the `setup()` method before the settings were populated.

adrianbj:
@ryancramerdesign - do you want Issues created to draw attention to PRs that are bug fixes?

ryancramerdesign:
Thanks Adrian, it looks like this one has been added.

adrianbj:
Hey Ryan - what do you mean by "**it looks like** this one has been added"? I didn't add `$this->setup();` - I simply moved it's location. If I look at the current dev branch, it's still in the old location which means it won't work. Maybe you have already changed in your local copy, but just wanted to make sure you understood what I actually did here and why it now works, but didn't before.

I know it's a simple line change, but it actually took quite a while to figure out why the InputfieldSelector wasn't respecting the "allowSystemTemplates" option.

adrianbj:
Hi @ryancramerdesign - I just checked the dev commit from today and `$this->setup();` is still on the wrong line and so still won't work. I don't want to hassle, but now that this PR has been closed (and I don't have the option to re-open it) I am worried it will be forgotten.

adrianbj:
@ryancramerdesign  - just checked 3.0.59 and it's still in the wrong spot. I was I could reopen this PR. I guess I'll submit an Issue report to bring this to your attention again :)

PR #52, opened 2017-01-25
Implement a way for modules to supply translations

LostKobrakai:
Module authors can ship translations of their modules by adding a label => relative path combo to their module info block. Paths must be relative to the modules root directory. The files should be exported csv files of just the module translations.

Example info block:
```php
'languages' => [
	'de_DE' => 'languages/de_DE.csv',
	'en' => 'languages/en.csv'
]
```

### DIFF
--- a/wire/modules/Process/ProcessModule/ProcessModule.module
+++ b/wire/modules/Process/ProcessModule/ProcessModule.module
@@ -21,6 +21,8 @@
 class ProcessModule extends Process {
 
 	public static function getModuleInfo() {
+		$languages = wire('languages');
+
 		return array(
 			'title' => __('Modules', __FILE__), // getModuleInfo title          
 			'summary' => __('List, edit or install/uninstall modules', __FILE__), // getModuleInfo summary
@@ -28,7 +30,7 @@ class ProcessModule extends Process {
 			'permanent' => true, 
 			'permission' => 'module-admin', 
 			'useNavJSON' => true,
-			'nav' => array(
+			'nav' => array_filter(array(
 				array(
 					'url' => '?site#tab_site_modules',
 					'label' => 'Site', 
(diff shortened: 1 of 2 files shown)


mikerockett:
I'm still not 100% knowledgable about languages in PW, but does everyone use the same language keys? Like what happens if someone just uses `de`?

wire/modules/Process/ProcessModule/ProcessModule.module
```diff
@@ -206,6 +215,11 @@ class ProcessModule extends Process {
 				// check that it can be installed NOW (i.e. all dependencies met)
 				if(!$this->wire('modules')->isInstallable($moduleName, true)) continue; 
 			}
+
+			if($translated) {
+				// exclude modules without translations present
+				if(!$info['languages'] || !is_array($info['languages']) || !count($info['languages'])) continue;
```
teppokoivula: Note: `!count($info['languages'])` is unnecessary since an empty array won't pass `!$info['languages']`. PHP converts an empty array to boolean `FALSE` and an array with items to boolean `TRUE`.

On the other hand you should first check if `isset($info['languages'])` or you might run into an `E_NOTICE` :)

LostKobrakai: Yeah, had that though yesterday evening as well. Might just use !empty(), which coverts the isset as well as count() issue.

LostKobrakai:
@mikerockett It's not automatically importing those translations. You can choose which translation file to import into which installed language. I just named my .csv's in the format of locales, so it's clear which kind of translation you can expect to import.

ryancramerdesign:
Thanks @LostKobrakai I have added this, with a few changes and additions. Biggest differences are that I went without the 'languages' property added to the moduleInfo since I thought the files in the /languages/ directory were already enough. I also left out the "Translated" navigation option in the modules nav, as I guess I thought there wasn't enough need for it, but will add it back if I'm wrong about that. Thanks for the nice addition.

PR #54, opened 2017-01-26
Fix bug when cropping images

esl51:
https://processwire.com/talk/topic/14574-image-cropping-not-working-in-backend/


ryancramerdesign:
I'm not sure I fully understand the issue here, and read through the thread linked above. I haven't seen any similar issues with cropping. But if the proposed adjustment fixes the issue you are seeing, it seems fine to add. The only thing is that we are intentionally using the min and max attributes to benefit from browser features provided by those attributes. Do you see any problem with us still having the min/max attributes present (in addition to the data-min and data-max attributes)?

esl51:
When submitting form, the browser can not focus on the hidden inputs (and at the moment of cropping this field is hidden, they are visible only when resizing) to show the message that the field is set to a value less than the minimum value (at this moment the value is 0, and it is not critical when cropping the image). Therefore, I propose to use the data-attributes here. This does not break the functional and preserves the possibility of HTML5-validating the visible fields.

ryancramerdesign:
> When submitting form, the browser can not focus on the hidden inputs (and at the moment of cropping this fields are hidden, they are visible only when resizing) to show the message that the field is set to a value less than the minimum value (at this moment the value is 0, and it is not critical when cropping the image).

Thanks for your reply. If I understand correctly, it sounds like we can't also use the HTML5 min/max attributes then –  because they are the source of the problem?

> This does not break the functional and preserves the possibility of HTML5-validating of visible fields.

Can you clarify on this point. How does it preserve HTML5-validation if we aren't using data attributes rather than the HTML5 non-data attributes? (or maybe HTML5 still recognizes data-min/data-max and I just don't realize it?). Thanks.

esl51:
> If I understand correctly, it sounds like we can't also use the HTML5 min/max attributes then – because they are the source of the problem?

Yes, the source of problem is hidden inputs with min- and max- attributes.

> Can you clarify on this point.

I mean other visible inputs, if they are exist.

Hurmeli:
Hi,

Has there been any movement on this problem? The bug still persists so cropping images does not work in backend which is a bummer.

adrianbj:
@Hurmeli - did you try the changes shown in the PR at the top of this thread - changing those `attr` to `data`

If you could let Ryan know if that works for you also, then maybe he can move forward on this?

Hurmeli:
I will try it tonight if I have time!

Hurmeli:
This indeed fixes the issue although there's still some strangish behavior.

Example:
My Images minimum size has been set to 2000x600. I upload a larger image and then use the crop tool to get the dimensions right.

1) Sometimes when using same image it ends up being 1999x600 px. I tried maybe dozen times and managed to only make this happen twice. Seems like rounding error of some sort and nothing too serious in most cases.

2) After cropping the image apperas resized to fit into the window and the numbers indicate something like 1300 x 390 or what ever. This feels confusing and to me feels like the cropping the image went wrong. Then when you touch the resize triangle in bottom right corner it pops up to the correct dimensions (2000x600).

ryancramerdesign:
> Has there been any movement on this problem? The bug still persists so cropping images does not work in backend which is a bummer.

I was never able to duplicate the issue. Cropping seems to work as intended when I test here. It's possible I missed something? Are there some specific steps necessary to observe the issue, or specific browser/version combination that it appears in?

Hurmeli:
It probably has to do with size of the image (and possibly the minimum size although I haven't tested this).

When you crop an image, you see it in the browser window. But the space for viewing the image is only around 1300 px wide. If it goes beyond this area, it gets scaled down so it fits it. If you try to crop it, the process will hang and nothing will happen (although sometimes rarely it does work).

Now if you click the icon to show the image in its original size, the cropping will not hang but it will instead crop the image wrong - seemingly based on the visible portion of the image instead of the whole image (you can still only see around 1300 px of the image, the rest will disappear outside the viewable area).

I have only tested the process with a largish image (2048 x 1500 px). Where the fields minimum size settings were 2000 x 600 px. Cropping failed every time. If you need more help with this I can write a step by step guide.

matjazpotocnik:
@esl51 @ryancramerdesign I think this can be closed.

ryancramerdesign:
@matjazpotocnik Thanks, I'll close it.

PR #55, opened 2017-01-30
added exception

steros:
throw new exception if field is not of expected type for better debugging.


matjazpotocnik:
@steros  Please fill CLA at https://processwire.com/about/license/cla/ --  specify the email address you want attributed for the PR (which should be the same as your GitHub email address).

PR #56, opened 2017-02-02
Update MarkupHTMLPurifier Module to load rel-Attribute values from Pr…

stereochrome:
…ocessPageEditLink

In ProcessPageEditLink possible values for rel attributes of links can be configured inside module config. The MarkupHTMLPurifier-Module removes all but nofollow. This Commit changes this behavior and loads possible values from ProcessPageEditLink. If this process does not exist, the codes falls back to original behavior.


billhemsley:
Note also PR#86, which deals with similar issues for target attributes with HTML Purifier.

ryancramerdesign:
Thanks @stereochrome I think we've got this part handled directly in InputfieldCKEditor now (though it wasn't when you submitted this PR). I'll close this one out for now, but if you think these changes would still be beneficial to make let me know and we can pull them in.

PR #57, opened 2017-02-06
Proposal: disable links

steros:
disable links that are either immediate parents or children of a editable field. thus you won't get redirected if you double click an editable region.
needs to be checked and minified cause I don't know what algorithm is used for minification.
I tested it with PW 3.0.42


mikerockett:
If the change is accepted, Ryan will manually add the change in a commit, using his minification process.

matjazpotocnik:
@steros  Please fill CLA at https://processwire.com/about/license/cla/ --  specify the email address you want attributed for the PR (which should be the same as your GitHub email address).

PR #58, opened 2017-02-07
Set cookie domain also for php session

derixithy:
This will remove the need to edit php.ini manual for use with cookie domains for the session() function (wire/wires cookie will be missing).


ryancramerdesign:
Thanks, this has been added.

PR #59, opened 2017-02-08
Change regex pattern to match closing edit tag

jofalk:
The closing HTML edit tag is not stripped out during page rendering if whitespace after edit is missing. The current pattern only matches:
`</edit >` but should `</edit>`. 
See e.g. source HTML of [Official Processwire Demonstration of front-end-editing](http://demo.processwire.com/regular/about/front-end-editor-demo/) where `</edit>` is not stripped out.


ryancramerdesign:
Thanks very much, that's a good find. I have added this (with a slight adjustment to the regex).

PR #61, opened 2017-02-14
closing </edit> tag not removed

steros:
You just shouldn't parse HTML with regex. This is another example why.
There are several other places too but unfortunately I don't have the time to fix them now :(


ryancramerdesign:
Thanks, we've got a fix in place for that particular regex which had an error in it.

steros:
Why fix the ugly regex and not replaced it with this more proper solution?

LostKobrakai:
Probably because DOMDocument is not a part of php ensured to be existing and it's easier to maintain one working solution for all people than rewriting a bunch of places to cause issues for some people.

PR #71, opened 2017-06-10
InputfieldIcons updates

rolandtoth:
- remove spans and use CSS to style icons instead inline styles
- do not re-init icons and events after toggling field table visibility
- clicking on a selected icon will clear icon selection
- long-click on an icon will show a larger preview with icon name

Preview:

https://processwire.com/talk/uploads/monthly_2017_05/icon-selector.gif.72801cfc143776217c6a2696bdd39a0e.gif


ryancramerdesign:
Thanks @rolandtoth great updates -- this PR has been added.

PR #74, opened 2017-06-20
Add event trigger for language tab clicks

rdunk:
Useful for initializing Javascript dependent Inputfields when switching language tabs.


ryancramerdesign:
@rdunk I meant to add an event like this when the dblclicklangtab event was added but looks like I didn't.  I have added your suggestion here with a different name for consistency with the existing event(s). Thanks.

PR #77, opened 2017-06-27
Make FieldtypeOptions detect single value usage

LostKobrakai:
Make FieldtypeOptions detect single value usage from the inputfield with markupValue() usage

### DIFF
--- a/wire/modules/Fieldtype/FieldtypeOptions/FieldtypeOptions.module
+++ b/wire/modules/Fieldtype/FieldtypeOptions/FieldtypeOptions.module
@@ -207,6 +207,10 @@ class FieldtypeOptions extends FieldtypeMulti implements Module {
 	 *
 	 */
 	public function ___markupValue(Page $page, Field $field, $value = null, $property = '') {
+		$interfaces = class_implements($field->inputfieldClass);
+		if(!in_array(InputfieldHasArrayValue::class, $interfaces) && $value instanceof WireArray) {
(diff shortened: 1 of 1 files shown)


wire/modules/Fieldtype/FieldtypeOptions/FieldtypeOptions.module
```diff
@@ -207,6 +207,10 @@ class FieldtypeOptions extends FieldtypeMulti implements Module {
 	 *
 	 */
 	public function ___markupValue(Page $page, Field $field, $value = null, $property = '') {
+		$interfaces = class_implements($field->inputfieldClass);
+		if(!in_array(InputfieldHasArrayValue::class, $interfaces) && $value instanceof WireArray) {
```
teppokoivula: Just an observation: this requires PHP >= 5.5.0. Is that OK?

LostKobrakai: I was worried about `class_implements`, but that's actually 5.1+. Didn't think of `::class`. Any better idea than hardcoding the string?

clsource: What about http://php.net/manual/en/function.get-class.php ?

ryancramerdesign:
Thanks @LostKobrakai this has been added (with minor changes).

PR #78, opened 2017-07-05
Fix z-index issue with selectize tags in image field

lesaff:
See commit comment for animated gif explanation of the issue.


matjazpotocnik:
@lesaff I'm not able to reproduce, but I'm running the latest PW version where some z-index issues with datefield and images were fixed and that might also fixed this one?

lesaff:
@matjazpotocnik I just upgraded the site to 3.0.95 and the z-index issue is still showing up.

matjazpotocnik:
Did you try with disabled cache? 
zindex

ryancramerdesign:
I can't seem to duplicate the issue that this PR fixes. Is this maybe already fixed or are there prerequisites necessary to duplicate? Thanks.

PR #80, opened 2017-08-04
Updated the texts for the english and german 404 page

weberhofer:
* The old texts were incorrect
* The second change in line "WireDatabaseBackup" has been created upon saving the file. Seems that a windows CR has been removed by the github editor.


jmartsch:
Texts look good to me 👍🏻

PR #81, opened 2017-08-10
Fix InputfieldWrapper::populateValues() when getting data from WireData objects

wanze:
Just a small "typo" :)


wanze:
It is actually not a typo, my bad. See: https://github.com/processwire/processwire-issues/issues/339

PR #88, opened 2017-09-14
Do not assume appendFiles to be in /site/templates

abdusco:
When trying to significantly change template router & render logic by hooking into Page::render, PW assumes append/prependFiles to be inside /site/templates directory and prepends /path/to/site/templates/ to any append/prepend file path I set. This happens when I try to modify page template to append a template file from a module's directory. 

```
/path/to/site/modules/MyModule/append.php => /path/to/site/templates/path/to/site/modules/MyModule/append.php
```

This change checks if given file path is absolute and within /site/ directory and if not, it defaults to templates directory and takes it from there.


abdusco:
The module I'm building is a [routing library](http://altorouter.com/). It hooks into Page::render and changes module filename so that every template now points to /site/templates/router.php where a router library can take over and handle the requests. This gives the all control of render routine to the developer.

PR #91, opened 2017-09-25
Respect noSizeAttr option in ProcessPageEditImageSelect

abdusco:
\ tags respect "Skip width attribute on image tags " setting in ProcessPageEditImageSelect module configuration.


matjazpotocnik:
@abdusco can this be closed?

PR #92, opened 2017-09-28
Improve getting of SVG image dimensions with $image->width / $image->height

adrianbj:


adrianbj:
I think this is really important when you want to add logic to your template files based on whether an SVG file is portrait or landscape. Currently this isn't possible if the SVG file doesn't have width and height tags set which is the case when exporting from Illustrator as a responsive image.

Please let me know if you have any questions.

adrianbj:
Any thoughts on this @ryancramerdesign - I like to keep my sites up to date with the latest core dev versions, but I keep having to add this in on each update. I'd just hate to see this forgotten about.

Thanks!

ryancramerdesign:
@adrianbj how does IMagick determine SVG width/height? It sounds like it must use some method other than the width/height attributes that we read from SimpleXML? Assuming that's the case, all sounds good, I agree we should add it. Do you know if it returns pixel dimensions or percentages?

adrianbj:
Hi @ryancramerdesign - if I understand correctly, it rasterizes the svg and determines how many pixels it takes up on the canvas. Even if an SVG doesn't have any dimensions specified, the paths within the SVG use co-ordinates to determine the position of nodes that are joined by lines etc. It is these co-ordinates that results in a "default" size for the SVG. iMagick returns pixel dimensions. If you look at the PR, you'll see that the new method returns the pixels from iMagick, but defaults back to 100% if it's not installed. 

One thing I considered doing for this PR, but didn't, was to actually move the xml attributes approach (which should be the default) for getting the dimensions into a new `svgSize()` method. This might be cleaner, but I took the approach of less modification to the current logic. Let me know if you'd like me to update this PR with that approach.

Thanks for getting this implemented.

ryancramerdesign:
That's pretty cool that it actually rasterizes it to determine dimensions.
I went ahead and added the PR, but in a little different way that sounds
maybe similar to what you mentioned for svgSize. I've tested the method out
here to make sure everything works, and it does. Though I don't have
Imagick installed at the moment, so haven't been able to confirm that part
of it works (though the code is identical to your PR), but please let me
know if you find anything doesn't work right. Thanks.

adrianbj:
All seems to work as expected - thanks.

PR #93, opened 2017-10-23
Edit multiple translation files in browser tabs

thuijzer:
See https://github.com/processwire/processwire-issues/issues/410

This change also sends the language_id via the URI so multiple browser tabs can be used to edit multiple languages of the same translation file.


ryancramerdesign:
Thanks for the PR. Just checked and it looks like this one was already added per processwire/processwire-issues#410, so will close this one out.

PR #94, opened 2017-11-05
SessionHandlerDB Option to disallow parallel sessions

kixe:
Added checkbox in mobile settings to restrict a user to only one single
session.
If this option is enabled after logging in, any sessions of this user
related to other devices and orphaned sessions (after removing the
session cookie) will be removed from the database.


ryancramerdesign:
Thanks, this seems like a nice addition, good idea. I have added your PR (with a few small changes).

PR #96, opened 2017-11-22
Support square brackets in field names

Notanotherdotcom:
Now supports showIf and requiredIf fields with names like skyscraper[0]name, skyscraper[1]name and so on in the same form.

This isn't useful in ProcessWire _at present_ since fieldnames can't contain brackets, but _is_ useful when I've been using PW as a framework/library and wanted to use inputfield dependencies in my frontend forms where I do have some array-based fieldnames.

Thought I'd submit it as it's a harmless change and might be useful in future or to someone else doing what I'm doing.

### DIFF
--- a/wire/templates-admin/scripts/inputfields.js
+++ b/wire/templates-admin/scripts/inputfields.js
@@ -177,6 +177,9 @@ function InputfieldDependencies($target) {
 
 		var $field = null;
 		var value;
+		
+		conditionField.replace('[', '\\[');
+        	conditionField.replace(']', '\\]');
 
 		consoleLog('getCheckboxFieldAndValue(see-next-line, ' + conditionField + ', ' + conditionSubfield + ')');
 		consoleLog(condition)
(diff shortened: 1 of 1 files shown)


Notanotherdotcom:
Apologies - example above should have been more like:

skyscraper[0][name]
skyscraper[1][name]

ryancramerdesign:
Pete, thanks, sounds like a worthwhile addition. But does this part work?
~~~~~~
conditionField.replace('[', '\\[');
conditionField.replace(']', '\\]');
~~~~~~
My understanding is that the replace() function returns a value, but doesn't modify the original string?

Notanotherdotcom:
You're absolutely right and they do nothing - I just tested it locally.

The bit that makes it work is actually the change on line 224 as that encapsulates the matched name in quotes so it knows that any square brackets within the name are part of the name string and not some sort of javascript sub-filter (my brain has drawn a blank trying to think of the correct term but hopefully you get what I mean :) ).

So basically it's just line 224 that needs changing for this to work.

ryancramerdesign:
Thanks Pete @Notanotherdotcom I have added this .

StanLindsey:
Close the branch if its complete? Get rid of PR noise?

StanLindsey:
Also - question - how come we don't merge the branch @ryancramerdesign and instead recreate it yourself? Wondering what the benefit is?

LostKobrakai:
@StanLindsey You notice that ryan's commit is different and includes the change of the minified js file? Using a merge instead of rebase/amending changes would mean those there would at least be two commits leaving one commit with a incorrect minified file. But I'm with you that closing PR's if they're done is the way to go for OOS. If there are still open questions it can just be reopened.

LostKobrakai:
@StanLindsey if you have a local workflow it's actually less work to simply merge stuff and amend any additional changes necessary. I highly doubt ryan is not initially merging PRs. I can certainly see why you feel ryan's way of handling things is not optimal (and in most places I probably think the same way), but imho we have to respect his decisions about how he does things. It has been criticised every now and then and I'm sure he's aware of the implications.

PR #98, opened 2017-12-05
Fix if server doesn't have mb_convert_encoding available.

adrianbj:
Maybe this should also be committed to https://github.com/erusev/parsedown-extra ?

### DIFF
--- a/wire/modules/Textformatter/TextformatterMarkdownExtra/parsedown-extra/ParsedownExtra.php
+++ b/wire/modules/Textformatter/TextformatterMarkdownExtra/parsedown-extra/ParsedownExtra.php
@@ -1,4 +1,4 @@
-<?php 
+<?php
 
 #
 #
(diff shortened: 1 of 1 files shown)


adrianbj:
There might be more of a conversation required here regarding the availability of `mbstring` in general. It looks like in most cases throughout the core you do a check, like:

```
$length = function_exists('mb_strlen') ? mb_strlen($value) : strlen($value);
``` 
or

```
function _strlen($s) {
    return function_exists('mb_strlen') ? mb_strlen($s, '8bit') : strlen($s);
}
```

These are great and just what we need, but there are places where these checks don't exist, eg in Sanitizer.php:

```
		for($n = 0; $n < mb_strlen($value); $n++) {
			$c = mb_substr($value, $n, 1);
```

Also, as well as parsedown-extra (as this PR deals with), there are also unchecked calls to several `mb_` functions in Punycode.php. 

So my question is - do you want to make `mbstring` a requirement for PW, or do you want to add conditionals with fallbacks in all places?

matjazpotocnik:
> do you want to make mbstring a requirement for PW

+1

ryancramerdesign:
I think we should probably make mbstring a requirement. It is kind of essential. How common is it not to have available? I have run into one case in the last year, and thought it was odd at the time that it wasn't present. While I'd prefer for mbstring to be a requirement, if it's commonly not available, we might be better off substituting our own versions.

PR #99, opened 2017-12-09
Add support for repeater field for InputfieldPage

karltaidev:
The dynamic change function is not supported in repeater field. This fix provides support for it.

### DIFF
--- a/wire/modules/Inputfield/InputfieldPage/InputfieldPage.js
+++ b/wire/modules/Inputfield/InputfieldPage/InputfieldPage.js
@@ -9,7 +9,7 @@ function initInputfieldPage($this) {
 
 	// support for dependent selects
 	$this.find(".findPagesSelector").each(function() {
-
+		var $repeater = $this.parents('.InputfieldRepeaterItem');
 		var $t = $(this);
 		var selector = $t.val();
 		// if there is no "=page." present in the selector, then this can't be a dependent select
@@ -30,11 +30,12 @@ function initInputfieldPage($this) {
 
 			var part = parts[n];
 			var name = part.replace('=page.', '');
-			var $inputfield = $('#Inputfield_' + name);
+			var $inputfield = $repeater.length === 0 ? $('#Inputfield_' + name) : $('#Inputfield_' + name + "_repeater" + $repeater.data('page'));
 			if($inputfield.length < 1) return;
 
 			// monitor changes to the dependency field
-			$inputfield.change(function() {
+			$inputfield.off('change');
+			$inputfield.on('change', function() {
 				var s = selector;
 				var v = $inputfield.val();
 				if(v == null) {


ryancramerdesign:
@karltdev I'd like to add this PR and potentially others of yours, and give you credit for it, but don't have a CLA from you. GitHub recommends we have a CLA for any contributions, are you able to complete it? https://processwire.com/about/license/cla/ -- if so, please fill it out and specify the email address you want attributed for the PR (which should be the same as your GitHub email address). 

I also wanted to ask about the part of the code that removes all change events via the off() method. Do you find that part is necessary? I was thinking it might be safer to remove a specific change event that the code requires, as there may be other change events active from unrelated modules (like ProDrafts). Or it's possible I've just misunderstood that part of the code. Thanks.

karltdev:
I am new to the open source world. Thanks for inviting me to fill out the CLA form and I have just done it.

> I was thinking it might be safer to remove a specific change event that the code requires, as there may be other change events active from unrelated modules (like ProDrafts). 

I agree. I did this because I did not know which change event(s) was/were influencing. Do you have any idea?

ryancramerdesign:
> I agree. I did this because I did not know which change event(s) was/were influencing. Do you have any idea?

@karltdev I'm not sure either, I usually try to leave existing events alone to be safe. That's why I'm wondering if the `off()` call is a necessary part of the logic in the PR? Is the purpose of it to avoid adding the same change event twice, or something else? The core InputfieldPage.js doesn't use an off() before adding the change event, so was curious. Thanks.

karltdev:
I just tried to review the code inside InputfieldPage.js, and found that the code has changed a bit. I tried not to use the `off()` call and it is still working. Below is the code I have tried.

```
        function initSelector($t) {
		var $repeater = $t.parents('.InputfieldRepeaterItem'); //edited
		var selector = $t.val();

		// if there is no "=page." present in the selector, then this can't be a dependent select
		if(selector.indexOf('=page.') == -1) return;

		var labelFieldName = $t.attr('data-label');
		var formatName = $t.attr('data-formatname');

		if(!labelFieldName.length) $labelFieldName = 'name';

		var $wrap = $t.parents(".InputfieldPage");
		var $select2 = $('select#' + $wrap.attr('id').replace(/^wrap_/, '')); // dependent select

		if($select2.length < 1) return;

		var parts = selector.match(/(=page.[_a-zA-Z0-9]+)/g);

		for(var n = 0; n < parts.length; n++) {

			var part = parts[n]; // page matching part of the selector
			var name = part.replace('=page.', '');
			var $select1 = $repeater.length === 0 ? $('#Inputfield_' + name) : $('#Inputfield_' + name + "_repeater" + $repeater.data('page')); //edited

			if($select1.length < 1) continue;

			// monitor changes to the dependency field
			$select1.change(function() {
				selectChanged($select1, $select2, selector, formatName, labelFieldName, part, true)
			});

			// determine if select2 needs to be populated
			if($select1.val() && !$select2.val() && $select2.children('option[value!=""]').length < 1) {
				// no options in $select2 select but primary $select1 has something selected…
				// it’s possible there are just no selectable options, but it’s also possible that 
				// this is a new page or the selects are newly added, so let’s find out…
				setTimeout(function() {
					selectChanged($select1, $select2, selector, formatName, labelFieldName, part, false)
				}, 100);
			}
		}
	}
```

ryancramerdesign:
Thanks @karltdev I have added this. I changed the `$t.parents(...)` to `$t.closest(...)` so that it would not match multiple repeater items in a nested repeater situation. Though let me know if you find that breaks anything intended.

Toutouwai:
A couple of comments on this change (not sure whether it's best to comment here or on the commit)...

1\. What about scenarios where you want the select inside the repeater item to be dependent on a select that's in the page open in Page Edit? That is, where the previous behaviour was desired. Maybe it should be configurable whether the "page" keyword in the Page Reference selector string refers to the page being edited or the repeater page.

2\. If I understand right, this change contradicts the resolution of an earlier issue: https://github.com/processwire/processwire-issues/issues/479
As a result of that issue a change was made to FieldtypePage::isValidPage() so that it would be consistent with what InputfieldPage considered a valid page for a Page Reference fields (where "page" was treated as the page being edited and not a repeater page).
This latest change seems to allow pages to be selected in Page Edit that are not savable because FieldtypePage::isValidPage() treats "page" as the page being edited: https://github.com/processwire/processwire/blob/04e2de28f057335979aa2a821fa4684d74303515/wire/modules/Fieldtype/FieldtypePage.module#L725-L728

A demonstration of the problem...

Page structure:


Suburb field config:


The "Suburb" field in the repeater saves when a suburb in Auckland is selected (because Auckland is selected in page being edited) but not when a suburb in a different city is saved:
selects

And if there is no "City" field in the page being edited then nothing can be saved in the "Suburb" field in the repeater:
selects2

karltdev:
Sorry it is my careless I did not test the save function as it was working before `FieldtypePage.module` fix. It looks like the conflict is hard to be resolved now as the repeater is no way to access the page itself when saving.

It looks like a configuration to create a scope for the repeater page as @Toutouwai  said is a possible solution?

```
while(wireInstanceOf($forPage, 'RepeaterPage') && ++$n < 10) { 
    /** @var RepeaterPage $forPage */ 
    if(!$forPage->scopeRepeater) $forPage = $forPage->getForPage(); 
} 
```

ryancramerdesign:
@Toutouwai @karltdev This is not a field scenario I've ever used before, but I think I understand what you guys are saying (good points Toutouwai too).  Though I don't think this case is common enough to warrant having it's own configuration setting. I'd rather just maintain the original behavior, while allowing for it to work in repeaters when the condition for the original behavior is not present, and the conditions for the repeater *are* present. I've pushed an update to this that I think does this, though I don't have the live test cases for this like it sounds like you guys do. Please let me know if you find this achieves what you were looking for, or if something different would be needed? Thanks.

PR #100, opened 2017-12-24
Code optimization and enhancements

mirzadabeer:
**Changes Made:
+++++++++++**

<strike>1] Added visibility(public/private/protected) for the constant.
	http://php.net/manual/en/language.oop5.constants.php#language.oop5.basic.class.this</strike>

<strike>2] Changed double quote to single quote wherever necessary.
	http://php.net/manual/en/language.types.string.php
	https://www.codeproject.com/Articles/988255/PHP-Single-Quotes-versus-Double-Quotes</strike>

<strike>3] Optimized the code by using Null Coalesce Operator instead of ternary operator
	http://php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce</strike>

4] Added expected exception in `CacheFile->save` function since we are using `$this->wire('files')` which tends to throw WireException.

5] Condition changed from `!in_array($page->process, array('ProcessLogin'))` to `$page->process != 'ProcessLogin'`.

Hopefully this helps, will be adding more changes to it after this PR merged.

Please let me know if you have any queries.

### DIFF
--- a/wire/core/CacheFile.php
+++ b/wire/core/CacheFile.php
@@ -66,7 +66,7 @@ class CacheFile extends Wire {
 			$this->globalExpireTime = @filemtime($this->globalExpireFile); 
 		}
 
-		$this->primaryID = $id ? $id : 'primaryID'; 
+		$this->primaryID = $id ? $id : 'primaryID';
 		$this->cacheTimeSeconds = (int) $cacheTimeSeconds; 
 	}
 
--- a/wire/core/CacheFile.php
+++ b/wire/core/CacheFile.php
@@ -268,7 +268,7 @@ class CacheFile extends Wire {
 	 *
 	 */
 	public function expireAll() {
-		$note = "The modification time of this file represents the time of the last usable cache file. " .
+		$note = "The modification time of this file represents the time of the last usable cache file. " . 
 			"Cache files older than this file are considered expired. " . date('m/d/y H:i:s');
 		@file_put_contents($this->globalExpireFile, $note, LOCK_EX); 
 	}
--- a/wire/core/CacheFile.php
+++ b/wire/core/CacheFile.php
@@ -18,14 +18,14 @@
  */
 class CacheFile extends Wire {
 
-	private const cacheFileExtension = '.cache';
-	private const globalExpireFilename = 'lastgood';
+	const cacheFileExtension = ".cache";
+	const globalExpireFilename = "lastgood";
 
 	/**
 	 * Max secondaryID cache files that will be allowed in a directory before it starts removing them. 
 	 *
 	 */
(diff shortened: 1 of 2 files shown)


processwired:
3: Note that the null coalescing operator is available as of PHP 7. I guess there are still a lot of older PHP versions installed out there, which would like to continue to run ProcessWire...

mikerockett:
Thanks for your contribution.

As @processwired indicated, PW supports older versions of PHP, so [3] cannot happen (as much as we’d like it to 😉).

Regarding [2] quotes: https://processwire.com/api/coding-style-guide/#8.1-single-vs-double-quotes. PW adheres to a style guide of its own. I may not be a fan of it, but it must be adhered to in the core, so this is also a no-go.

processwired:
One more thing: 
`$note = "The modification time of this file represents the time of the last usable cache file. " . "Cache files older than this file are considered expired. " . date('m/d/y H:i:s');`
is actually faster than the "optimization" you proposed: 
`$note = 'The modification time of this file represents the time of the last usable cache file. ' . 'Cache files older than this file are considered expired. ' . date('m/d/y H:i:s');`

processwired:
Got to revert this a bit - I now did some performance comparisons with that strings. It shows the same processing time for me, considering fault tolerance. If any optimization, than get rid off the first unnecessary concatenation between "...file" and "Cache...", but this is more for the eyes than for speed.

As soon as a variable gets included in the double quoted string, it slightly performs better than the single quoted alternative. But again, no considerable potential for more speed, though. It's more a matter of personal taste or code convention.

mirzadabeer:
@processwired @mikerockett Thank you for your comments, I agree to all your points, excluding all these updated the only change(Number 4 & 5).

Please consider.

matjazpotocnik:
@mirzadabeer @ryancramerdesign As I can see there is only one change:

`!in_array($page->process, array('ProcessLogin'))` is replaced by `$page->process != 'ProcessLogin'`

Is there really any benefit or significant performance gain?

ryancramerdesign:
@matjazpotocnik While I doubt it makes any difference here, in_array() can sometimes be slow, depending on the number of items being compared. I'm guessing this line of code originally had more than one item, or that more were going to be added, but never were. If I were coding it new I would write `"$page->process" !== "ProcessLogin"` which would probably be the fastest method. But for existing code, I don't get into micro optimization unless doing some other related refactoring nearby.

ryancramerdesign:
@mirzadabeer I added the change to admin.php. I wasn't able to confirm the method you added the `@throws` phpdoc to does throw a WireException, but if you find it does, please let me know how to reproduce and we can add that too.

PR #101, opened 2018-01-09
Update ProcessPageSearch.module

Renobird:
Add support for "^= starts with" and "$= ends with" operators. Appears to resolve the issue where "starts with" and "ends with" are non-functional in Page Auto Complete fields. I'm not seeing any other impacts, however, there may be a reason these were omitted that I'm not aware of.

https://github.com/processwire/processwire-issues/issues/468


ryancramerdesign:
Thanks @Renobird I'll close this since we added these from your issue report awhile back.

PR #102, opened 2018-01-12
Add FieldtypeIcon

BitPoet:
Add FieldtypeIcon complementing InputfieldIcon so this can be used in
templates/pages like other input fields. Basically, this is a stripped-down FieldtypeText. This module implements feature request https://github.com/processwire/processwire-requests/issues/128.

### DIFF
--- a/wire/modules/Fieldtype/FieldtypeIcon.module
+++ b/wire/modules/Fieldtype/FieldtypeIcon.module
@@ -1,0 +1,72 @@
+<?php namespace ProcessWire;
+
+/**
+ * ProcessWire Icon Fieldtype
+ *
+ * Complements the core icon inputfield so it can be used in templates/pages
+ *
+ */
+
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
I think I'd rather leave this one as a 3rd party module since I'm guessing this wouldn't meet the 30% usage threshold that we try to stick to for new core modules. But it would be a great one to have available, if you don't mind releasing in the modules directory, if it doesn't already exist there.

teppokoivula:
This is something that I end up using in almost all of my projects, but of course that alone won't meet that 30% threshold either.

Honestly I would've assumed that this is a very common need, and I've seen a number of use cases solving this in a different (less elegant) manner, such as text field with icon names, Page references, etc. Still, without any real data this is just guesswork from both sides :)

kixe:
I use this one https://github.com/kixe/PageIcon

PR #107, opened 2018-05-06
trigger pw-panel-closed event

BernhardBaumrock:
Hey @ryancramerdesign,

could you please add this trigger to pw-panels? I use panels a lot for my Datatables to modify existing pages and want to reload the grid whenever a panel was closed.

I'm not sure about the minified version... I created it with https://javascript-minifier.com/ - is there a recommended way how to do PRs on javascript files that also have a minified version?

Thanks

PS: Should I have taken pw:dev as target for the merge? Sorry... not yet familiar with that...


LostKobrakai:
> is there a recommended way how to do PRs on javascript files that also have a minified version?

afaik ryan does all his merges on the local environment, so just do the PR on the unminified file and the minified one should be generated after the merge.

ryancramerdesign:
No problem, I have added this so that it triggers your pw-panel-closed event, as well as a corresponding pw-panel-opened event.

BernhardBaumrock:
thanks!

PR #108, opened 2018-05-08
pass value of confirm dialog to the callback

BernhardBaumrock:
vex uses a callback for handling the confirm dialog. that callback receives true when OK was clicked and false when the user clicked CANCEL. this value needs to be passed to the callback in both cases. without this fix it only calls the callback on success (OK).

### DIFF
--- a/wire/modules/Page/PageFrontEdit/PageFrontEdit.module
+++ b/wire/modules/Page/PageFrontEdit/PageFrontEdit.module
@@ -307,7 +307,10 @@ class PageFrontEdit extends WireData implements Module {
 
 		// if given field name doesn't map to a custom field, delegate the request to $page instead
 		$renderField = false;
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
The intention for ProcessWire.confirm() is that it doesn't call the callback unless the user confirms. So if I added this PR, it looks like it would break all the existing usages of ProcessWire.confirm(), and in a bad way (since it is often used to confirm a destructive action). I understand the desire to have this though. What about if we put it in a different function, like ProcessWire.approve() (or 'agree' or 'affirm')? OR: update the confirm() function to accept a 2nd callback that represents the false state? We could do either of those without breaking existing usages.

BernhardBaumrock:
Thx, ryan. I see the problems. I think a second callback would be a great solution!

PR #110, opened 2018-05-30
Expand language tabs button fixes

rolandtoth:
`cfg.liActiveClass` seems to be empty in Reno/Default themes so line 112 evaluates to `$ul.find('.').addClass(...)`. This is causing a JavaScript error and makes the Expand language tab button not working at all.


ryancramerdesign:
Thanks for the fixes! I have added your two PRs for LanguageTabs.

rolandtoth:
Working great, thanks!

PR #113, opened 2018-06-22
Allow use of "id" as automatic page name

Notanotherdotcom:
Could probably really do with allowing any field or combination of fields, but granted that's a much bigger job :)

### DIFF
--- a/wire/core/PagesEditor.php
+++ b/wire/core/PagesEditor.php
@@ -398,6 +398,9 @@ class PagesEditor extends Wire {
 				if(strlen($page->title)) $pageName = $page->title;
 					else $pageName = $this->untitledPageName;
 
+			} elseif($format == 'id') {
+				$pageName = $page->id;
+
 			} else if(!ctype_alnum($format) && !preg_match('/^[-_a-zA-Z0-9]+$/', $format)) {
 				// it is a date format
 				$pageName = date($format);


wire/core/PagesEditor.php
```diff
@@ -398,6 +398,9 @@ public function setupPageName(Page $page, array $options = array()) {
 				if(strlen($page->title)) $pageName = $page->title;
 					else $pageName = $this->untitledPageName;
 
+			} elseif($format == 'id') {
```
StanLindsey: `elseif` > `else if`

adrianbj: While I also prefer `elseif` it seems that @ryancramerdesign prefers `else if`. Do you have a reason for why one is actually better than the other?

lesaff: `elseif` is the recommended way per PSR-2
https://www.php-fig.org/psr/psr-2/#51-if-elseif-else

teppokoivula: While referring to PSR-2 is a reasonable argument, we haven't adopted that for ProcessWire. That being said, our own style guide recommends `else if` instead of `elseif` as well: https://processwire.com/api/coding-style-guide/#5.1-if-elseif-else :)

StanLindsey:
Aha. I actually meant change `elseif` to `else if`. I meant that as an arrow instead of a greater than sign. Lol

jmartsch:
There is a module for adding support of using dynamic values (as ID) in the page name:
[ProcessSetupPageName](http://modules.processwire.com/modules/process-setup-page-name/)
However, I would also like to see support for this in the core.

PR #114, opened 2018-07-07
Add REST PUT Request Body to $input

StanLindsey:
This adds the PUT Request Body onto the input field. Allowing you to do the following:
AJAX in developer console:
```javascript
fetch('website.com/post/', {
    method: 'POST',
    headers: {
      'Accept': 'application/json',
      'Content-Type': 'application/json'
    },
    body: JSON.stringify({a: 1, b: 'Textual content'})
  }).then(function(response) {
    return response.json();
  })
  .then(function(myJson) {
    console.log(myJson);
  });
```

post template:
```php
header("Content-Type: application/json; charset=UTF-8");
header("Access-Control-Allow-Methods: POST");

// Get the request body
$data = $input->get('requestBody');

// Respond with it
echo json_encode([
	'originalData' => $data,
])
```

### DIFF
--- a/wire/core/WireInput.php
+++ b/wire/core/WireInput.php
@@ -120,6 +120,7 @@ class WireInput extends Wire {
 	 *
 	 */
 	public function get($key = '') {
+		if($key === 'requestBody') return $this->requestBody();
 		if(is_null($this->getVars)) {
 			$this->getVars = $this->wire(new WireInputData($_GET));
 			$this->getVars->offsetUnset('it');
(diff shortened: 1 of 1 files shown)


wire/core/WireInput.php
```diff
@@ -120,6 +120,7 @@ public function __construct() {
 	 *
 	 */
 	public function get($key = '') {
+		if($key === 'requestBody') return $this->requestBody();
```
StanLindsey: Not sure if this is the best place for this - but didn't feel getVars was correct either.

StanLindsey:
@ryancramerdesign

PR #116, opened 2018-07-26
add $page->httpsUrl property+method

BernhardBaumrock:
I want to force HTTPS for live environments but not for dev. We already have the option of forcing https via htaccess but this also forces https on dev. We also have the relatively new $config->noHTTPS variable, but then I need to set all templates to "https" manually.

I want a global config/redirect based on the environment and with this PR and this oneliner in ready.php it would easily be possible:
```
if(!$config->https AND !$config->noHTTPS) $session->redirect($page->httpsUrl);
```

What do you think?

### DIFF
--- a/wire/core/Page.php
+++ b/wire/core/Page.php
@@ -34,6 +34,7 @@
  * @property string $url The page’s URL path from the server's document root
  * @property array $urls All URLs the page is accessible from, whether current, former and multi-language. #pw-advanced
  * @property string $httpUrl Same as $page->url, except includes scheme (http or https) and hostname.
+ * @property string $httpsUrl Same as $page->httpUrl but with HTTPS forced on
  * @property Page|string|int $parent The parent Page object or a NullPage if there is no parent. For assignment, you may also use the parent path (string) or id (integer). #pw-group-traversal
  * @property Page|null $parentPrevious Previous parent, if parent was changed. #pw-group-traversal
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
@BernhardBaumrock The `http` in the method name `httpUrl` refers to “Hypertext transfer protocol" rather than http vs. https (port 80 vs. port 443 ssl). So adding an httpsUrl method might cause confusion about the meaning of the httpUrl method, since the 'http' portion is not meant to refer directly to a specific http scheme/port but rather to hypertext transfer protocol in general. In addition, I think that other than specific cases, it might be problematic to write code referring to a specific URL scheme/port since it is likely to change between dev, staging and production. Granted we've got `$config->noHTTPS` to get around that, but coding for specific things like scheme or hostname feels like a practice that probably shouldn't be facilitated with dedicated methods. I think your idea is good, but instead maybe it fits better as options to existing methods. I've added support for this to force use of https (or another scheme) in the httpUrl: 
~~~~~
$page->httpUrl([ 'scheme' => 'https' ]); 
~~~~~
it also works with the regular url() method since forcing a scheme also requires a hostname:
~~~~~
$page->url([ 'scheme' => 'https' ]); 
~~~~~

BernhardBaumrock:
Thx! I needed this quite often. Eg when sending links via mail where I want to make sure that users land on the HTTPS version of that page (eg when having an input form).

PR #117, opened 2018-08-03
added a more verbose debug info

horst-n:
It is usefull to have all possible debug infos for images in one central point.

### DIFF
--- a/wire/core/Pageimage.php
+++ b/wire/core/Pageimage.php
@@ -1539,14 +1539,14 @@ class Pageimage extends Pagefile {
 	}
 
 	/**
-	 * Optionally set options to be picked up from the __debugInfo method
+	 * Get DebugInfo, optionally with individual options array 
 	 * 
-	 * @return pageimage object
+	 * @return array
 	 * 
 	 */
-	public function setDebugOptions($options = array()) {
+	public function getDebugInfo($options = array()) {
 		$this->debugOptions = $options;
-		return $this;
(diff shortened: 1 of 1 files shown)


horst-n:
blue is the default output, green is additional in verbose mode:

pw-image-debug-info-verbose

adrianbj:
Hey @horst-n - this looks really great. I am curious though - is there a reason why my foundSupportingEngines key is empty in this example? Is it just because nothing special is needed for a JPG, or is this a bug?

image

adrianbj:
Nevermind - I just read this comment on the `getEngines` method:

> Note that the returned value excludes the default engine (ImageSizerEngineGD).

horst-n:
@ryancramerdesign Ryan, also this is not an issue, it would be really helpful to have this enhanced debugInfo integrated into the current dev branch, before it becomes the new stable.
It doesn't affect any other method or param, only enhances the __debugInfo() method with useful informations. This would be really helpful for debug and support cases, if someone encounters a problem and ask in the forums for help, we can get a good insight / overview in his/her current images system by simply asking to show as that info.

horst-n:
@ryancramerdesign I've reworked it a bit so that we can load any individual options array that may get passed into the imagesizer. This way we can get the 100% exact information for every size() call.
It also respects and reflects individual options like "forceNew" or "forceEngine".
Look at the following example:
screen_enhanced_individual_debuginfo_pageimages

PR #118, opened 2018-08-14
fix for issue #628

horst-n:
changed from the old behaviour to now return a small variation that fit into aspect ratio of the target dimensions.

@ryancramerdesign I have tested this thouroughful and it seems to work with all cases:
- No dimension to small; One dimension to small; Two dimension to small;
- Requesting a landscape and source is: landscape; portrait; square;
- Requesting a portrait and source is: landscape; portrait; square;
- Requesting a square and source is: landscape; portrait; square;
All with enabled / disabled $config->imageSizerOptions['cropping']


horst-n:
@ryancramerdesign 
Hi Ryan, within the last days, where I have used the new default bevaviour, I encountered that the change was a bit more obstrusive as I thought first. But now, after ading support for sharpening, all aspects are covered.

I also cleaned up some parts in the ImageSizerEngineGD.php for better reading and understanding.

ryancramerdesign:
Thanks @horst-n this is great! I am adding this today.

horst-n:
Thanks Ryan!

PR #120, opened 2018-08-28
Bugfix null values

thuijzer:
Fixes the PHP error `Warning: A non-numeric value encountered in /wire/modules/Fieldtype/FieldtypeFloat.module on line 107`


matjazpotocnik:
@thuijzer @ryancramerdesign this is already fixed so it could be closed? https://github.com/processwire/processwire/blob/a7c85f11d9a7e0c85a9a8b9d172500428d38887c/wire/modules/Fieldtype/FieldtypeFloat.module#L135

PR #121, opened 2018-08-29
make pw-content-head-buttons clickable on smaller screens

tobaco:
#pw-content-head-buttons wasn't clickable on smaller screens because of an overlap from the headline


rolandtoth:
@tobaco This was caused by the AdminOnSteroids module (and was fixed similarly to your PR). Could you confirm and perhaps close this?

PR #122, opened 2018-09-03
Declare cookies as secure if on HTTPS page

posixpascal:
During one of our monthly security scans we noticed that this cookie does not include the "secure" flag. It's not a biggie at all but  we had to make this change in our ProcessWire core installation to prevent the security tool from reporting it. 

Can we merge this into the core so we can keep our code base upgradable?

### DIFF
--- a/wire/modules/Inputfield/InputfieldImage/InputfieldImage.js
+++ b/wire/modules/Inputfield/InputfieldImage/InputfieldImage.js
@@ -1419,7 +1419,11 @@ function InputfieldImage($) {
 		if(!name.length || typeof value == "undefined") return;
 		if(data[name][property] == value) return; // if already set with same value, exit now
 		data[name][property] = value; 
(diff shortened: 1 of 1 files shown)


posixpascal:
Hey @ryancramerdesign, sorry to bother you with this.

Are there any updates on this PR? 
We can't upgrade our instance as security scans would mark our backend as insecure if the secure cookie flag is missing.

ryancramerdesign:
Hi @posixpascal there isn't anything in this particular cookie that would need to be secure. All that it does is remember the last selected type of image layout (grid vs list). I would consider flagging these kinds of cookies to be a false positive from the security scanner. Though I also don't think there's any harm in using the secure option either (other than for people that might switch between http and https). I'll add this update or something similar on the next dev branch version. Thanks.

netcarver:
@posixpascal This has been manually adopted into the codebase now. Are you able to close this PR now?

PR #124, opened 2018-10-18
Add hookable method to PageImage variation creation

karltdev:
Reference request: [https://github.com/processwire/processwire-requests/issues/223](https://github.com/processwire/processwire-requests/issues/223)

### DIFF
--- a/wire/core/Pageimage.php
+++ b/wire/core/Pageimage.php
@@ -728,8 +728,8 @@ class Pageimage extends Pagefile {
 						}
 					}
 					
-					if($sizer->resize($width, $height) && @rename($filenameUnvalidated, $filenameFinal)) {
-						$this->wire('files')->chmod($filenameFinal);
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
@karltdev Sounds like this would be worthwhile to add, but since the added method doesn't actually create a variation, I'm wondering if we would be better off with a hookable method named `___createdVariation()` (past tense) that is called after the variation has been created and renamed? ... Or a hookable method named `___createVariationReady()` that is called before the variation is created? (i.e. before the resize method is called?)

matjazpotocnik:
@karltdev, thoughts on Ryan's comments?

karltdev:
I agree with the name ___createdVariation()

ryancramerdesign:
Thanks @karltdev this has been added.

matjazpotocnik:
@karltdev can you close this PR?

PR #126, opened 2018-10-23
uk-disabled class removed from language tabs

tiefenb:
Regarding: https://github.com/processwire/processwire-issues/issues/734


tiefenb:
okay. I see this is now fixed with this commit: https://github.com/processwire/processwire/commit/e65c2870995c0922ce4e60e5779815337739b102

PR #127, opened 2018-11-03
Expose actual user and role/s through ProcessWire.config javascript object for AdminThemeUikit

jmartsch:
This PR would resolve my request https://github.com/processwire/processwire-requests/issues/241

### DIFF
--- a/wire/modules/AdminTheme/AdminThemeUikit/init.php
+++ b/wire/modules/AdminTheme/AdminThemeUikit/init.php
@@ -40,6 +40,13 @@ $ukGridWidths = array(
 $config->set('inputfieldColumnWidthSpacing', 0); 
 $config->js('ukGridWidths', $ukGridWidths);
 
+$roles = array();
+foreach ($this->user->roles as $role) {
+	array_push($roles, $role->name);
+}
+$config->js('roles', $roles);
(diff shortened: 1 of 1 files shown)


wire/modules/AdminTheme/AdminThemeUikit/init.php
```diff
@@ -40,6 +40,13 @@
 $config->set('inputfieldColumnWidthSpacing', 0); 
 $config->js('ukGridWidths', $ukGridWidths);
 
+$roles = array();
+foreach ($this->user->roles as $role) {
+	array_push($roles, $role->name);
+}
+$config->js('roles', $roles);
```
rolandtoth: Why not just `$config->js('roles', $this->user->roles->name())`?

jmartsch: Because it returns an array. I thought about using the roles as an additional css class on the body element. I think your code is better if you plan only for using it from the javascript side, but then we would need an additional script to add the classes to the body.

Hmmm, really don't know whats best.

jmartsch: Yes, your code is much simpler

PR #129, opened 2018-11-08
Write ajax uploaded files to disk in chunks in WireUpload

BitPoet:
Read ajax uploaded files from php://input and write them to disk in chunks of 4MB in WireUpload to preservice memory as outlined in processwire/processwire-requests#233


BitPoet:
Incorporated to dev through commit 93a9747657e55c8e19c6aea9284657173cc734d4

PR #130, opened 2018-11-29
add comments regarding FieldtypeDecimal

BernhardBaumrock:


LostKobrakai:
I really like what you did there, but would tackle it a bit differently. Instead of promoting the decimal fieldtype directly we need to inform the user about all the potential backdraws floats have first. Only after that's done the user can make an informed decision and this is where we can link to the decimal fieldtype.

Currently I don't feel like a user not aware of the difference between floats and decimals could make much sense about why this is there without doing the work and clicking all the links.

I also don't feel like the note is needed in that place. Why and how floats work differently then one might expect can be part of external information we link to. Important is that we warn about the issues of floats, explanation can happen elsewhere.

LostKobrakai:
http://0.30000000000000004.com/ might be a good resource so link to

BernhardBaumrock:
I'd vote for a human readable blogpost :D And add your link there ;)

teppokoivula:
I do have to disagree with that last one. The explanation in the PHP docs is indeed good, but I don't think it's particularly user-friendly, or understandable for most users. Couple of quotes:

> The size of a float is platform-dependent, although a maximum of approximately 1.8e308 with a precision of roughly 14 decimal digits is a common value (the 64 bit IEEE format).

> PHP typically uses the IEEE 754 double precision format, which will give a maximum relative error due to rounding in the order of 1.11e-16. Non elementary arithmetic operations may give larger errors, and, of course, error propagation must be considered when several operations are compounded.

That's exactly the kind of complexity that, in my opinion, a system such as ProcessWire is supposed to handle for the user. One shouldn't have to understand all that just to reliably calculate decimal values. (And yes, I'm using decimal here, because I'd argue that this is the typical real world use case for our float field, even though it's not intended for that.)

PHP docs link to floating-point-gui.de for the "simple" (human-readable, if you ask me) version of the explanation, and I think we should probably start with that – or, preferably, write our own version, which roughly covers the important points, and explains where to use floats, and where *not* to use them :)

ryancramerdesign:
@BernhardBaumrock Since the decimal option is now in the core, are there any parts of this PR that you think should still be added? Thanks.

BernhardBaumrock:
no thx :)

PR #131, opened 2018-12-03
Write better notice message

jmartsch:
if pre-selection can not be populated, because the old one is very confusing.


wire/modules/Fieldtype/FieldtypeOptions/FieldtypeOptions.module
```diff
@@ -109,7 +109,7 @@ class FieldtypeOptions extends FieldtypeMulti implements Module {
 			} else if($this->wire('process') != 'ProcessField' && !wireCount($value)) {
 				$this->warning(
 					$field->getLabel() . " ($field->name): " . 
-					$this->_('Configured pre-selection not populated since value is not always required. Please correct this field configuration.') 
+					$this->_('Your pre-selection could not be populated. To enable the pre-selection, you must activate the "required" field option for this field. You can find it under the input tab in the fields settings.')
```
teppokoivula: Minor notes on the language: "field option for this field" could make more sense as "option for this field", and "the input tab in the fields settings" should probably be "the input tab in the field\'s settings".

PR #132, opened 2018-12-03
use better non-technical message

jmartsch:


ryancramerdesign:
As far as I can tell this is a duplicate of #133 (or rather 133 a duplicate of this). Since 133 has been added I'll close this one.

PR #133, opened 2019-01-26
use better non-technical message

jmartsch:
if pre-selection can not be populated, because the old one is very confusing.


ryancramerdesign:
Thanks @jmartsch this has been added

PR #134, opened 2019-02-22
WireHttp.php - Allow user specified cURL options

chriswthomson:
Allow the user to specify cURL options by passing them as an array through $options["curl"]. 
For example, $http->post("https://api.endpoint.com/", $data, ["curl" => [CURLOPT_USERPWD => "AnAPIKeyExample123"]]).


ryancramerdesign:
Thanks @chriswthomson this has been added. Note it's been changed slightly to use `$options['curl']['setopt']` rather than `$options['curl']` per the commit message.

PR #137, opened 2019-03-11
make list of all sanitizers hookable

BernhardBaumrock:
Robin created a new sanitizer and I checked if it shows up in $sanitizer->testAll() results, but it didn't. I made a little tweak in his module that only works if we have a hookable getAll() method in the sanitizers class.

See https://processwire.com/talk/topic/21018-sanitizer-transliterate/

### DIFF
--- a/wire/core/Sanitizer.php
+++ b/wire/core/Sanitizer.php
@@ -3567,7 +3567,22 @@ class Sanitizer extends Wire {
 	 * 
 	 */
 	public function testAll($value) {
-		$sanitizers = array(
+		$sanitizers = $this->getAll();
+		sort($sanitizers);
+		$results = array();
+		foreach($sanitizers as $method) {
+			$results[$method] = $this->$method($value);
+		}
+		return $results;
+	}
+
+	/**
+	 * Get all available sanitizers
+	 *
+	 * @return void
+	 */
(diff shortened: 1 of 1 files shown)


wire/core/Sanitizer.php
```diff
@@ -3567,7 +3567,22 @@ public function maxBytes($value, $maxBytes = 128) {
...
+		foreach($sanitizers as $method) {
+			$results[$method] = $this->$method($value);
+		}
+		return $results;
+	}
+
+	/**
+	 * Get all available sanitizers
+	 *
+	 * @return void
```
teppokoivula: This isn't quite right – should be `@return array`.

BernhardBaumrock: thx :)

netcarver:
@BernhardBaumrock As this is an issue, perhaps you could raise an new issue on processwire-issues and link to this PR. I can then mark the issue as containing a suggested fix.

BernhardBaumrock:
Hi @netcarver I've added a comment on this existing issue: https://github.com/processwire/processwire-issues/issues/85 and just saw that it is already closed... Can you reopen it?

BernhardBaumrock:
Yes, thx :)

PR #138, opened 2019-03-11
add support for WireArray->count(selector)

BernhardBaumrock:


wire/core/WireArray.php
```diff
@@ -1679,13 +1679,20 @@ public function getIterator() {
...
 	 * 
+	 * You can also pass a selector:
+	 * 
+	 * ~~~~~
+	 * $qty = $items->count("foo=bar");
+	 * ~~~~~
+	 * 
 	 * #pw-group-retrieval
 	 * 
 	 * @return int
```
teppokoivula: This DocBlock should have a param definition, perhaps something along the lines of `@param string|null $selector Optional selector string to match items against.`

ryancramerdesign:
@BernhardBaumrock I think this might make it technically incompatible with PHP’s Countable interface that we are implementing (https://www.php.net/manual/en/class.countable.php) even if not functionally incompatible. Seems like depending on the PHP version, it might not like violating the interface (which adding, removing or changing arguments does), though I could be wrong. Perhaps an alternate method name for this might be safer? (findCount, getCount, something like that?)

PR #144, opened 2019-05-28
Bugfix: $input->requestMethod() does not return anything when HTTP method is PATCH

MrSnoozles:
When getting the current requestMethod with `$input->requestMethod()` an empty string is returned when the method is `PATCH`. This pull request fixes this bug.


netcarver:
@MrSnoozles As this has been fixed by Ryan, are you able to close this PR? (I don't have any admin rights on this repo so I can't.)

PR #145, opened 2019-06-11
Update PagefileExtra.php

TomS-:
Support for browser based fallback of WebP support, eliminating the need for extra markup such as the picture element or modification of the htaccess file. This allows you to use the srcset and sizes element. 

Edit: For those of you who use UIKit this is great because you can use the built in image component that doesn't support picture elements.

### DIFF
--- a/wire/core/PagefileExtra.php
+++ b/wire/core/PagefileExtra.php
@@ -140,21 +140,22 @@ class PagefileExtra extends WireData {
 	 * 
 	 */
 	public function url($fallback = true) {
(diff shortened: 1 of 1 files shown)


horst-n:
Using something like ```if(strpos($_SERVER['HTTP_ACCEPT'] ...``` does eliminate every usage of cache(s)!! This is a really bad idea.

TomS-:
Oops. Haha, yeah @horst-n I didn't think of that.

PR #146, opened 2019-07-21
Make it possible to activate inline CKEditor instances via keyboard

teppokoivula:
This PR should make inline CKEditor instances more usable for keyboard users:

- `.InputfieldCKEditorInline` container divs gets `tabindex=0`, which makes it possible to focus them with keyboard (tab).
- `ckeInlineMouseoverEvent` was changed to `ckeInlineActivateEvent` and applied to the focus event, as well as mouseover. Name change isn't really necessary, but seemed to make sense.
- In the `ckeInlineActivateEvent` event there's a bit of focus-related code. The purpose of this is to make sure that the cursor moves to the editable area. At least on Chrome this was necessary; otherwise you'd see editor tools, but had to move focus to another element and back again to be able to actually edit the content.

Overall it seems that these changes shouldn't really break anything else, but I haven't had the chance to test them a whole lot. Also wasn't sure how you've usually minified JS – for this PR I used uglifyjs with `--compress --mangle`.

Feel free to modify any parts, but overall I feel that this feature – assuming that it works as intended – would be a nice addition from accessibility point of view. Thanks for considering! :)


ryancramerdesign:
Thanks @teppokoivula this has been added

PR #147, opened 2019-08-03
Update the JSON format in the module info JSON example

teppokoivula:
This PR would resolve the issue mentioned in https://github.com/processwire/processwire-issues/issues/952.


teppokoivula:
This was fixed in f22302aaaf0c3ee2ea77392c80098bf999e326a8.

PR #149, opened 2019-08-20
Added default non-www redirect

webmanufaktur:
Non-WWW redirect to keep all possible options someone might need.


netcarver:
@webmanufaktur Thanks for this fix. If this is an issue for you, please could you open a new issue on processwire-issues and link to this PR. I'll then label it as having a suggested fix.

webmanufaktur:
There is no need for this PR anymore. A non-www set was added already.
See Line 325 in htaccess.txt/.htaccess

PR #151, opened 2019-10-11
Show Textdomain in Language Translator

tiefenb:
I have many translatable PHP Files ordered in different directories but with the same filename.

I.e.:
* /foo/bar/baz/translations.php
* /bar/foo/baz/translations.php
*  /baz/bar/foo/translations.php

Only the basename of the file is shown in the current language translator.
Showing the textdomain would help me in the translation process.


tiefenb:
Preview:

image

ryancramerdesign:
Thanks @tiefenb I have added this (or rather something close to it).

PR #153, opened 2019-10-17
Update PaginatedArray.php

Schnapsum:
Fix for last page getPaginationString when count mod limit != 0(e.g. with 15 pages on limit 10)
currently: P1: 1 to 10 of 15, P2: 11 to 20 of 20
fixed: P1: 1 to 10 of 15, P2: 11 to 15 of 15


matjazpotocnik:
@Schnapsum A CLA is a prerequisite for PRs, and this is because GitHub recommends we have a CLA for any contributions. Are you able to complete it? https://processwire.com/about/license/cla/ -- if so, please fill it out and specify the email address you want attributed for the PR (which should be the same as your GitHub email address).

PR #155, opened 2019-11-02
Fix PHP 7.4.0-RC4 Warning

tiefenb:
I get this error on some PW admin pages: Notice: Trying to access array offset on value of type int in /var/www/html/wire/core/PagesLoader.php on line 138


tiefenb:
Fixed in: https://github.com/processwire/processwire/commit/76943ac192d194c79a0b0ecc49f11c41665ab362

PR #160, opened 2020-02-10
enable file metadata in repeater context

gerritvanaaken:
even when file/image was freshly created via Ajax call. fix https://github.com/processwire/processwire-issues/issues/1070

### DIFF
--- a/wire/modules/Inputfield/InputfieldFile/InputfieldFile.module
+++ b/wire/modules/Inputfield/InputfieldFile/InputfieldFile.module
@@ -293,8 +293,9 @@ class InputfieldFile extends Inputfield implements InputfieldItemList, Inputfiel
 	 * @return string
 	 * 
 	 */
-	protected function pagefileId(Pagefile $pagefile) {
-		return $this->name . "_" . $pagefile->hash; 
+	protected function pagefileId(Pagefile $pagefile, $repeaterstring = '') {
+
+		return $this->name . "_" . $repeaterstring . $pagefile->hash; 
 	}
 
 	/**
@@ -1227,7 +1228,15 @@ class InputfieldFile extends Inputfield implements InputfieldItemList, Inputfiel
 
 		/** @var Page $page */
 		$page = $pagefiles->getFieldsPage();
-		$id = $item ? ('_' . $this->pagefileId($item)) : '';
+
+
+		$contextpage = wire('pages')->get(wire('input')->id);
+		if (in_array(get_class($contextpage), ['ProcessWire\RepeaterPage', 'ProcessWire\RepeaterMatrixPage'])) {
+			$repeaterstring = "repeater".$contextpage->id."_";
+		} else {
+			$repeaterstring = "";
+		}
+		$id = $item ? ('_' . $this->pagefileId($item, $repeaterstring)) : '';
 		
 		$inputfields = $this->itemFieldgroup->getPageInputfields($page, $id, '', false); 
 		if(!$inputfields) return false;


gerritvanaaken:
need some proper coding, this is some ugly hack, but seems to work on two independent installation (one clean, one in production)

daun:
I'm using this on multiple installations as well and haven't had any issues. Would love to see this merged.

ryancramerdesign:
@gerritvanaaken I'd like to add something like this PR and give you credit for it, but don't have a CLA from you. GitHub recommends we have a CLA for any contributions, are you able to complete it? https://processwire.com/about/license/cla/ -- if so, please fill it out and specify the email address you want attributed for the PR (which should be the same as your GitHub email address).

gerritvanaaken:
Thanks Ryan. I just accepted the terms, of course. Let’s get this bug fixed for the next major version :-)

ryancramerdesign:
Thanks @gerritvanaaken I have added your PR. I made a change so that it gets the context page from data already available in the Pagefiles object rather than from `$input->id`. They are the same as far as I can tell, but I'm sure my setup is also different from yours. Can you confirm the committed version still resolves the issue you were seeing there? Thanks.

gerritvanaaken:
Thanks Ryan, but it still does not work in my setup. I investigated a bit, but with no elegant solution. 

- In my setup, I have a regular repeater within(!) a matrix repeater. (more than one repeater dimension.)
- The "context page" that I fetch via  $input->id must always be the root context page (the one that is currently edited), then the $id will be generated correctly.
- your solution: $pagefiles->getPage() gets only the direct parent as far as I understand this. Works well in one dimension, not in more dimensions :-(

Maybe there is a more elegant way to get the "root context page". The rest of your code restructuring works fine, it’s only line 1332, where you get the context page.

ryancramerdesign:
@gerritvanaaken Thanks, it looks like I misunderstood what was needed here. But we do have a way we can get that information. Below is the existing code with one line added, indicated with the comment. Can you try adding it to see if that enables it to work? 
~~~~~
if($page && method_exists($page, 'getForPage') && wireInstanceOf($page, 'RepeaterPage')) {
  $page = $page->getForPageRoot(); // ADD THIS LINE
  $id = '_' . $this->pagefileId($item, "repeater{$page->id}_");
} else {
  $id = '_' . $this->pagefileId($item);
}
~~~~~
You mentioned the "root context page (the one being edited)". Though I think it's possible it might need the page that owns the repeater item, rather than the nested root of them. Though I might not be remembering that right. But if you find the above doesn't work, try replacing the added line with this:
~~~~~
$page = $page->getForPage(); // ADD THIS LINE
~~~~~

gerritvanaaken:
@ryancramerdesign Still no luck :-(

- `$pagefiles->getPage()->getForPage();` does "not exist or is not allowed in this context", possibly because of Ajax-created repeater item
- `$pagefiles->getPage()->getForPageRoot();` delivers always the page id of the page being edited in the visible browser window, which _seems_ to be right, but does not work in Ajax context either.
- My hacky `input->id` always works in Ajax and non-Ajax context. It delivers the repeater subpage id when used in Ajax context, and the id of the currently edited visible page in the browser window.

This is my test code:

``````
if($item) {
	$page = $pagefiles->getPage()->getForPageRoot();
	wire('log')->save('debug', 'Ryan "getForPageRoot": ' . $page->id);
	$page = wire('pages')->get(wire('input')->id);
	wire('log')->save('debug', 'Gerrit "input": ' . $page->id);
	if(wireInstanceOf($page, 'RepeaterPage')) {
		$id = '_' . $this->pagefileId($item, "repeater{$page->id}_");
	} else {
		$id = '_' . $this->pagefileId($item);
	}
} else {
	$id = '';
}
``````

In the screenshot attached the commented result of the logfile!

Bildschirmfoto 2021-05-21 um 23 19 20

ryancramerdesign:
@gerritvanaaken I've pushed another update to this and was wondering if you could try it out? I've been reluctant to use the id from input because it is from user input and thus could be anything, plus there's no guarantee that we're in a page editor. I've tried to solve those by first checking that the current process is a page editor of some kind or another (ProcessPageEdit, ProcessUser, ProcessProfile, ListerPro, etc.) and then getting the page (and id) from that WirePageEditor instance, rather than from user input. It seems to be working in my testing here, but I'm not certain my setup/environment is the same as yours. Please let me know what you find when you get a chance. Thanks.

gerritvanaaken:
@ryancramerdesign This works instantly on my test setup! Thank you for giving this additional thoughts!

ryancramerdesign:
@gerritvanaaken Great! Glad that works. Thanks for testing it out

PR #161, opened 2020-02-26
allow config variable for multi-dimensional $input arrays

porl:
Added a config setting to allow for multidimensional arrays in $input as discussed here: 

https://processwire.com/talk/topic/691-wireinput-only-allows-one-dimensional-arrays-why/?tab=comments

Obviously it should be set to false by default (as I have done in the core/config.php file).

A better solution might be to allow the choosing of depth but I needed something quick for a site I am working on and this works well enough.

### DIFF
--- a/wire/config.php
+++ b/wire/config.php
@@ -927,6 +927,16 @@ $config->cookieOptions = array(
 	'fallback' => true, // If set cookie fails (perhaps due to output already sent), attempt to set at beginning of next request? (default=true)
 );
 
+/**
+ * Allow Multi-Dimensional Arrays in $input API var?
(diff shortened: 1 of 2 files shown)


porl:
Gah!! Made a typo (extra closing parenthesis). Fixed in commit 
23f2970

ryancramerdesign:
Thanks @porl this has been added, but with a little bit different implementation.

porl:
Thanks! I ended up using the _POST variable directly for the site I was working on (so updating PW didn't undo my changes) but this looks like a much cleaner solution for the future!

PR #162, opened 2020-02-26
Allows users to view their own page, even if they’re not allowed to edit it.

JanRomero:
Hi,

I’ve noticed that users with associated templates cannot view their own user page in the front-end unless they have the permissions `user-admin` or `profile-edit`. I figure these permissions are about editing, not viewing, so this is what I came up with to honor `$options['viewable']`. Hope I got it right.

Thanks


Lazerproof:
Also have noticed this issue but in the context of multiple templates or parents for users. @JanRomero maybe you can open an issue and link this PR to it&

ryancramerdesign:
Thanks @JanRomero something similar was added this week, so I'll close this one out.

PR #163, opened 2020-02-26
this old issue is still open, maybe a conditional check can help?

horst-n:
otherwise we should strip the utf8_encode and simply save the $pal as is.
We mainly do need the value of $pal for the check in line 108, and not so much the export in the info array.


ryancramerdesign:
Thanks @horst-n this has been added

PR #165, opened 2020-03-18
make CKE plugins path hookable

BernhardBaumrock:
This little addition makes it possible to add custom CKEditor plugin paths. Paths can then be added easily via hook:

```php
$wire->addHookAfter("InputfieldCKEditor::getPluginPaths", function(HookEvent $event) {
  $paths = $event->return;
  $paths[] = $this->config->paths->assets . "test/plugins/";
  $event->return = $paths;
});
```

This has the benefit that plugin authors can place plugin code in their dedicated module folder and have it properly controlled via GIT.

Example of how it has to be done without this modification: https://github.com/Toutouwai/headingscase

>  Copy the "headingscase" folder to /site/modules/InputfieldCKEditor/plugins/

This is not only tedious/unnecessary, more than that it is problematic. What if the file needs an update? What about bugfixes?

Thx.

### DIFF
--- a/wire/core/WireFileTools.php
+++ b/wire/core/WireFileTools.php
@@ -1347,15 +1347,21 @@ class WireFileTools extends Wire {
 	/**
 	 * Convert local/server disk path to URL
 	 * 
+	 * http parameter can be one of the following:
+	 * 		false: relative url
+	 * 		true: http/https (current setting of $config->https)
+	 * 		https: force https
+	 * 		http: force http
+	 * 
 	 * @param string $path Server disk path to dir or file that maps somewhere within web document root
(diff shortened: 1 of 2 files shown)


ryancramerdesign:
Thanks @BernhardBaumrock I have added support for this, but ended up doing it in a different way because I didn't want a module have to be autoload just to affect plugins, and thought maybe better if it adds an extra plugin dir as part of its install/uninstall. See the `InputfieldCKEditor::addExtraPluginsDir()` method for details on usage and please let me know if you run into any issues with it. Thanks.

BernhardBaumrock:
Thx :)

PR #166, opened 2020-03-18
simplify callable autoload param test

pine3ree:
just shorter code


ryancramerdesign:
@pine3ree I think your change here accomplishes shortening the 2 lines of code into 1 line very well, but admittedly I sometimes like a little bit more verbose code just for my own readability (so long as it's not adding overhead). I'd prefer to stick with the longer version for now, but it's a good suggestion. Thanks.

PR #169, opened 2020-03-25
fix inputfields not toggling on header clicks

BernhardBaumrock:
This PR fixes an old and annoying bug:

Before:
bug

Its a little hard to see, but the following happens:

* opened inputfield: arrow blinks
* collapsed inputfield: toggles correctly
* ajax: opens correctly, blinks instead of closing

After:
fix

With this fix in place all inputfields toggle correctly on the first click on the header.

### DIFF
--- a/wire/templates-admin/scripts/inputfields.js
+++ b/wire/templates-admin/scripts/inputfields.js
@@ -1880,8 +1880,6 @@ function InputfieldStates($target) {
 				});
 				if(isTab) {
 					$header.effect('highlight', 500);
-				} else {
-					$header.click();
 				}
 			}, 500);
 		}, 'html');
@@ -1934,6 +1932,7 @@ function InputfieldStates($target) {
 		var isCollapsed = $li.hasClass("InputfieldStateCollapsed"); 
 		var wasCollapsed = $li.hasClass("InputfieldStateWasCollapsed");
 		var duration = 100;
+		var isAjax = $li.hasClass('collapsed10') || $li.hasClass('collapsed11');
 	
 		if(!$li.length) return;
 		if($li.hasClass('InputfieldAjaxLoading')) return false;
@@ -1943,24 +1942,16 @@ function InputfieldStates($target) {
 			if(typeof data.duration != "undefined") duration = data.duration;
 		}
 
-		if(isCollapsed && ($li.hasClass('collapsed10') || $li.hasClass('collapsed11'))) {
+		if(isCollapsed && isAjax) {
 			if(InputfieldStateAjaxClick($li)) return false;
 		}
-			
+		
 		if(isCollapsed || wasCollapsed || isIcon) {
 			$li.addClass('InputfieldStateWasCollapsed'); // this class only used here
 			Inputfields.toggle($li, null, duration);
 		} else {
-			// Inputfield not collapsible unless toggle icon clicked, so pulsate the toggle icon and focus any inputs
-			if(typeof jQuery.ui != 'undefined') {
-				var color1 = $icon.css('color');
-				var color2 = $li.children('.InputfieldHeader, .ui-widget-header').css('color'); 
-				$icon.css('color', color2);
-				$icon.effect('pulsate', 300, function () {
-					$icon.css('color', color1);
-				});
-			}
-			Inputfields.focus($li);
+			// Inputfield not collapsible unless toggle icon clicked
+			$icon.click();
 		}
 
 		return false;


BernhardBaumrock:
One thing to mention is that the class `InputfieldStateWasCollapsed` remains on the inputfield even after toggling. I don't know what it is for, but it does not seem to have any effect...

BernhardBaumrock:
I'm toggling many items at once like this:
```js
  $(document).on('click', '.icon-fold', function(e) {
    var $rbb = $(e.target).closest('ul.Inputfields');
    var fields = $rbb.find('> li:not(.rbb-fold)');
    Inputfields.close(fields);
    e.preventDefault();
  });
```

Surprisingly this worked when clicked the FIRST time, but for all following clicks this script failed. Turned out that Inputfields.toggle() was meant to take only one element (strange that it worked nevertheless!).

teppokoivula:
@BernhardBaumrock, if I got this right, I believe this behaviour was discussed in in processwire/processwire-issues/issues/900, and even before that in processwire/processwire-issues/issues/855.

@ryancramerdesign's explanation makes sense considering that the label really should focus the field, and nothing else — and, in this regard, I think the "collapsed inputfield: toggles correctly" part is actually the only one that's misbehaving: Ryan mentions that at that point it's expected behaviour (first click opened the field, so it makes sense for the second one to close it)... but I think it's a problem for a11y :)

BernhardBaumrock:
Thx @teppokoivula, that was an interesting read!

Though I have to say that I really disagree! @ryancramerdesign I totally understand your point of view and especially from a technical perspective what you are saying makes sense. BUT... please - just for a minute - put yourself in the position of the user: This is **REALLY** bad UX!

Imagine you are visiting this page:

image

> You might think, "oh, there are lots of fields. What is behind those dots beside `Downloads`". So you click on the headline and the field opens. Great! Let's close it again. Works!
> 
> Ok, what's that? `Suchindex`... let's click on it. What's that little rotating thing? Ah... seems to be loading... Opens! Great! Let's close it again. Click. Nothing. Click. Nothing.

I guess 99% of the users wouldn't even realize the little flashing arrow on the far right. And I've made my browser smaller for this demo - on wider screens it gets even worse!
 
> Ok, let's close all those fields above the bodycopy to get some space. Click. Nothing. Click. Nothing... Ah... again I have to click this nasty little icon...

I bet that happens many many times every day all around the world (**including myself - after years of using ProcessWire!**).

Again - I understand your technical thoughts about the issue - but once more, step into the shoes of the client:

> Ok, I clicked on the headline just like I did before. It does not behave like before... That has a reason for sure. Let's inspect the code... Ah, this headline is actually a `label` of an HTML inputfield. Yeah, clicking on labels should focus the corresponding inputfield - great that the field didn't toggle as I expected, now the field is focused as it should be and I can click on the little icon... ähm..... ah, here... perfect.

### This will not happen! 💯 

---

And finally, if that short case story was not enough, I also have some technical arguments against what you said @ryancramerdesign 

* You state that this has accessibility benefits, but the inputfield elements are not accessible via keyboard one way or another (or did I miss something?)
* You state that when a user has opened the inputfield once, the assumption is that on the next click of the header he wants to toggle the field closed. Ok, so what about Inputfields with state collapsedYesAjax? The search index field described in the story above is such a field. It opens correctly, but it does not close when clicking the header. That's even a bug taking when applying your logic ;) And further more, that's actually not an Inputfield, which is the next point:
* InputfieldMarkup is not an inputfield and therefore has no focus(). Nevertheless an opened InputfieldMarkup does NOT collapse when one clicks on the header. This is absolutely counter-intuitive and in that case imho even technically wrong.

Inputfields are a great way of structuring page edit screens or presenting data in admin pages. I'm using them all over in my process modules. You can build great dashboards in the matter of seconds:

```php
// executeDashboard()
...
$form->add([
  'type' => 'markup',
  'label' => 'Mailbox (2)',
  'value' => 'You have 2 unread messages',
  'columnWidth' => 50,
],[
  'type' => 'markup',
  'label' => 'Website Statistics',
  'value' => 'You have had 1 visitor in the last 24 hours',
  'columnWidth' => 50,
]);
return $form->render();
```

There's really no reason why those page elements should not toggle when clicking on the field header!

Thanks in advance, I think this will be a great improvement for many many users out there and make them even happier when using our products 👍 



PS: This is exactly the behaviour all the modern and professional frameworks use for their toggling elements:

Uikit:
uikit toggle

Bootstrap:
bootstrap accordion

webmanufaktur:
This post and the support thread reminded me that I kind of successfully ignored this weird behaviour for a very long time now.

I understand why this behaviour is intended to be this way but... not my clients. A while back I toggled fields on a lot of templates to reduce the complexity and declutter those pages. But more often clients were confused in some cases therefore I stopped using toggled-fields and still today I try to avoid using them. That made my life easier as clients weren't confused anymore.

From a user perspective the existing behaviour is hard to understand. Especially for those users that are not that keen in doing "online stuff". It's in some cases even hard for clients to _tick_ that small arrow. Especially older users.

Just my two cents and a +1 for @BernhardBaumrock 's fix.

BernhardBaumrock:
I've added a quickfix for this in my pwquickfixes module: https://github.com/BernhardBaumrock/PwQuickFixes/commit/74c1378d0274bd6a99147bac597490d0113c2633

```js
$(document).on('click', '.InputfieldHeader', function() {
	$(this).closest('li.Inputfield').addClass('InputfieldStateWasCollapsed');
});
```

ryancramerdesign:
@BernhardBaumrock Thanks for the PR. It sounds like you feel more strongly about the behavior of these than I do, so I'm happy to add your suggested PR here as an optional setting, and have done so. But I've made it an option that you can select in the AdminThemeUikit module settings, as I'd rather people decide they want that behavior (for now) rather than changing it for them. The reason is that the label as an open/close toggle only makes sense on forms that have some collapsed fields already. For instance, front-end forms (like those in FormBuilder, LoginRegisterPro, among others) rarely use the option at all (or in many cases don't support it), so it could be confusing to have field labels showing and hiding the inputs. The admin environment is a safer bet though, so I've left it as something where we can later change the default for the admin theme if/when it makes sense after marinating awhile For now, you can select this option which uses your PR in Modules > Configure > AdminThemeUikit > Forms + Inputs.

BernhardBaumrock:
Thx!!! :) I know that I'm not the only one that will like this update @webmanufaktur 👍

matjazpotocnik:
@ryancramerdesign I would also like to thank you for the effort put into this PR (and all others too). For consistency and a cherry on the cake, it would be nice to have this behaviour in all admin themes.

PR #173, opened 2020-05-14
Update InputfieldPageName

markusthomasde:
Fix to correct replacment for german umlauts


teppokoivula:
@markusthomasde, note that this array provides just the default conversion values — you can override them in module config if you need something else. Current defaults are the expected ones for many languages, so changing them globally is, in my opinion, not feasible.

This might fix the situation for German umlauts, but it would simultaneously break things for existing sites using, for an example, Finnish or Swedish :)

markusthomasde:
@ryancramerdesign I can live with that. In a non multilingual installation this is also correct, but what if I have to use for example German and Swedish as languages? 

Then I have to choose one version and another language is rewritten incorrectly.

Maybe the URL rewrite should also become multilingual.

BernhardBaumrock:
While I agree that this would be nice, you can always use hooks to generate custom page names on page save:
```php
$wire->addHookAfter("Pages::saveReady(template=foo)", function($event) {
  $page = $event->arguments(0);
  $replace = ["ä"=>"ae", "ö"=>"oe", "ü"=>"ue"];
  $name = str_replace(array_keys($replace), array_values($replace), $page->title);
  $page->name = $event->sanitizer->pageName($name);
});
```
This should also work for multilang. You'd just need to set `name$lang` property: https://processwire.com/talk/topic/4720-how-to-set-page-name-in-different-languages-through-the-api/

PR #175, opened 2020-07-14
Page::changed receives same $old and $new value for objects

MrSnoozles:
The issue is described in https://github.com/processwire/processwire-issues/issues/1177

When the field value is an object (e.g. MultiLanguage fields) Page::changed receives the same value for $old and $new and is called twice. This change fixes the behaviour.


MrSnoozles:
Closed in favor of updated #186

PR #178, opened 2020-08-11
Update WireInput.php

theGC:
array reverse resets the index to 0 based, correct `$get` to accommodate for this when retrieving from the end of `urlSegments`


ryancramerdesign:
Thanks @theGC it looks like your fix is already present in the WireInput file, so I'll close this one.

PR #180, opened 2020-09-03
Allow root page to be selected in "Page List Select" when current page has children

kevinvermeersch:
Fix for issue => https://github.com/processwire/processwire-issues/issues/1151


kevinvermeersch:
A little background on the issue: The ProcessPageEdit.module overwrites the default "showRootPage" option when the current page has children.

https://github.com/processwire/processwire/blob/d8945198f4a6a60dab23bd0462e8a6285369dcb9/wire/modules/Process/ProcessPageEdit/ProcessPageEdit.module#L1102

matjazpotocnik:
@kevinvermeersch is this fixed in https://github.com/processwire/processwire/commit/ee4c46a442f09cf81f9e1800c5fdfcd1176f538a ?

kevinvermeersch:
@matjazpotocnik Yes this is fixed

PR #182, opened 2020-10-11
Add empty composer.lock file

strugee:
It doesn't do anything, but Heroku deploys complain if it's not there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,6 @@ upgrade.php
 .sass-cache/*
 *.css.map
 wire/modules/AdminTheme/AdminThemeDefault/styles/sass/*.css
-composer.lock
 *.min.min.js
 wire/modules/Inputfield/InputfieldDatetime/timepicker/i18n/*.min.js
 wire/modules/Inputfield/InputfieldDatetime/jquery-ui-timepicker-addon.min.js
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,20 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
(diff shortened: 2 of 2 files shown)


ryancramerdesign:
@strugee Since this is not something that ProcessWire itself would use or need, and is apparently to resolve an issue with a web host, I'm not sure this is something we should add but will leave it open for further feedback to see what others say. Thanks.

strugee:
Yeah, I went a bit further down this path and basically decided Heroku was a bad fit for deploying ProcessWire. If you want to merge be my guest, but honestly I don't think it's really necessary anymore.

ryancramerdesign:
Thanks @strugee I'll close this one then. Though when I next hear from someone using Heroku with PW I'll be sure to mention this to them that they should add a composer.lock file like the one you have here.

strugee:
Sure thing! If someone mentions using Heroku though my advice would be to tell them to use something else. Heroku just isn't really a good tool for the job; the whole model is basically predicated around the idea that the app runs in a container with an ephemeral filesystem.

I suspect you could figure something out with a Heroku addon to persist parts of the filesystem, but you'd have to wire a bunch of stuff up yourself and you'd have a bad time making sure you properly separated your ProcessWire site's code from its data - something ProcessWire is not really designed for. And if you don't do that, you can't actually take advantage of much of what makes Heroku operationally attractive in the first place.

ryancramerdesign:
Thanks AJ, this is good to know. I'll make note of these considerations
with Heroku.

strugee:
Any time. If someone comes asking about it too feel free to ping me.

PR #183, opened 2020-10-16
Make ProcessLanguageTranslator.module translatable

yuters:
The file ProcessLanguageTranslator.module can't be translated because it shows in its comments the text to add in the comments for making a file not translatable, so it skips itself. This check fixes this hilarious bug, without having to remove the comment explaining how to do this.


ryancramerdesign:
Thanks @yuters this has been added

PR #185, opened 2021-01-22
fix missing bottom margin on InputfieldImage

jmartsch:
The description field or the InputfieldImageEdit__core field has no bottom margin. 
So if you have custom image fields, it looks very ugly and disctracting.
This PR fixes the margin.

This is how it looks without the fix
image

And the fixed version
image


ryancramerdesign:
Thanks @jmartsch I have added this

PR #187, opened 2021-04-26
Add cookie SameSite support

pine3ree:
Update php `setcookie()` calls in order to support the now required SameSite attribute.

For PHP<7.3 a php bug is exploited and the SameSite attribute is injected into the Path parameter
For PHP>=7.3 the new setcookie signature is used.

Since `SameSite=None` is not widely supported by older browser, a default `Lax` value is used, but that can be changed via configuration both for standard and session cookies.

PS
I just noticed that my editor removed every trailing whitespace encountered and replaced tabs with 4 spaces...I'll try to revert that and force-push in order to have a cleaner diff...

### DIFF
--- a/wire/config.php
+++ b/wire/config.php
@@ -4,18 +4,18 @@
  * ProcessWire Configuration File
  *
  * Configuration options for ProcessWire
- * 
+ *
  * To override any of these options, copy the option you want to modify to
  * /site/config.php and adjust as you see fit. Options in /site/config.php
- * override those in this file. 
- * 
- * You may also make up your own configuration options by assigning them 
- * in /site/config.php 
- * 
+ * override those in this file.
+ *
+ * You may also make up your own configuration options by assigning them
+ * in /site/config.php
+ *
  * ProcessWire 3.x, Copyright 2016 by Ryan Cramer
  * https://processwire.com
  *
- * 
+ *
  * TABLE OF CONTENTS
  * ===============================
  * 1. System modes
@@ -23,15 +23,15 @@
  * 3. Session
  * 4. Template files
  * 5. Files and assets
- * 6. HTTP and input 
+ * 6. HTTP and input
  * 7. Database
  * 8. Modules
  * 9. Misc
- * 10. Runtime 
+ * 10. Runtime
  * 11. System
- * 
+ *
  * @var Config $config
- * 
+ *
  */
 
 if(!defined("PROCESSWIRE")) die();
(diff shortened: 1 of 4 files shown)


ryancramerdesign:
@pine3ree Thanks for the PR. I'm wondering about the PHP<7.3 bug you mentioned -- despite googling a bit, I can't seem to find any information about it, and php.net docs for `setcookie` and `session.cookie_samesite` doesn't seem to say anything about it either. The only thing I can find is something about a bug in PHP<7.3 when using `None` as the SameSite value. Can you point me to more information about this PHP bug? I want to make sure I fully understand it while adding the PR.

pine3ree:
Hello @ryancramerdesign ,

the bug is described here : https://bugs.php.net/bug.php?id=69948 and fixed in php 7.3.7 https://www.php.net/ChangeLog-7.php, but for php>=7.3.0 we already have the new signature for setcookie;

I found a lot of info on stackoverflow about how to leverage this bug to send the samesite attribute for lower versions of php. It seems to be a very common solution. I implemented variants of it for different CMSes and for websites still running on php 5.3. Another way (the first I used) is to use `header()` directly after building the set-cookie response header, but we still have to make the session send the attribute as well. Injecting the samecookie attribute the simplest common way to do it both in `setcookie`, `session_set_cookie_params` /`ini_set('session.....')` and `session_start($options)`. For php >= 7.3 samesite is supported in each mentioned functions.

I have read (MDN I believe) that `None` can cause issues for older browsers, and as of now it requires setting the Secure attribute as well. So a cookie set with `SameSite=None; Secure` will not be sent when accessing the site with `http` scheme.

Lax is considered the default value (MDN), so I set it to be the default in my pr. It works with non https scheme and if some other site links (GET req) to my pw website, stored cookies are sent by the browser when visiting (POST redirects from other sites are not with Lax, since POST request are usually state changing requests)

kind regards

ryancramerdesign:
@pine3ree Great, thanks I think I understand now. So it's not the PHP bug that's a problem to work around, but rather a bug that we can use to our benefit in setting the SameSite value. Very creative. :)

pine3ree:
Yes, exactly @ryancramerdesign,

...and the funniest thing is the version coincidence.... 

I applied similar patches on many website (also pw 2.5 and pw 2.7) without any issue, but please test it thoroughlly yourself. pw is your kid!!! :-)

ryancramerdesign:
Thanks @pine3ree great addition, this has been added on the dev branch.

PR #189, opened 2021-05-20
Add LESS support for AdminThemeUikit

BernhardBaumrock:
This PR makes it possible to easily modify any aspect of the PW backend theme using LESS variables.

See the readme here: https://github.com/processwire/processwire/blob/8a7f934a419fe2bf2463623bf7ac32dd49364513/wire/modules/AdminTheme/AdminThemeUikit/readme.md


wire/modules/AdminTheme/AdminThemeUikit/AdminThemeUikit.module
```diff
@@ -1107,7 +1107,9 @@ class AdminThemeUikit extends AdminThemeFramework implements Module, Configurabl
 			
 			// check if we need to recompile the theme
 			$recompile = false;
-			if(self::dev OR $config->recompile) $recompile = true;
+			if(self::dev OR $config->recompile OR !file_exists($compiled)) {
```
teppokoivula, line 1115: Seems that this PR uses AND/OR instead of &&/||. Not a big deal and Ryan likely adapts these things anyway, it's just something that kind of stuck out :)

BernhardBaumrock: Is there something bad about AND / OR ? :) Or is it just PW code style?

teppokoivula: Nothing bad specifically, it's just that when our entire codebase (AFAIK) uses another convention, these stuck out a bit. `and` and `or` have different precedence compared to `&&` and `||` (basically both `&&` and `||` are considered more important) but that doesn't matter a whole lot as long as one doesn't accidentally combine these (i.e. use `and` and `||`) :)

I'm not sure if our coding standard says anything about this, mind you. Likely not — it just felt weird when single piece of code (module) used both 🤷‍♂️

BernhardBaumrock: Interesting. Sure, I agree that it's nice to stick to one version. Sorry for that. Though the linked comment does not apply to my PHP environment:
img

teppokoivula: Again, this is a very small detail, and to be honest for me it was mostly the style part.

What you're debugging there is not exactly the same thing: the example I linked to includes `=`, which has a higher precedence than `and`, which in turn means that the variable gets a different value than it would if you replaced `and` with `&&`. This is not directly related to the code here, just an example of a type of gotcha that one can run into if they assume that operators in question are the same, while in reality they are not *exactly* the same :P

BernhardBaumrock: Ah, thx teppo now I get the difference :) I was a little confused about that example and thought maybe that's something that changed from php 5 to 7 or whatnot. Makes sense now 👍

PR #190, opened 2021-05-22
FieldtypeMulti: orderby `sort` only if schema supports it

pine3ree:
otherwise a SQL `column not found` exception is thrown with custom fieldtypes extending from FieldtypeMulti and not needing a manual sorting feature.

### DIFF
--- a/wire/core/FieldtypeMulti.php
+++ b/wire/core/FieldtypeMulti.php
@@ -520,8 +520,10 @@ abstract class FieldtypeMulti extends Fieldtype {
 		
 		if(empty($orderByCols)) {
 			// if there are no orderByCols defined, pagination & sorting not supported
-			// default sort for FieldtypeMulti fields is by column 'sort'
-			$query->orderby("$table.sort");
+			// default sort for FieldtypeMulti fields is by column 'sort', if schema supports it
+			if (isset($schema['sort'])) {
+				$query->orderby("$table.sort");
+			}
 
 		} else {
 			// one or more orderByCols is defined, enabling sorting and potential pagination


ryancramerdesign:
@pine3ree Thanks, I can add this. But I'm curious, have you come across a FieldtypeMulti that doesn't have a 'sort' column? (I might be forgetting an obvious example, just can't recall).

ryancramerdesign:
@pine3ree Actually I don't think this will work here (?) because `$schema` will not ever have a `sort` index in it since it is a trimmed schema excluding columns common to all FieldtypeMulti (such as 'pages_id' and 'sort'). We could maintain a separate `$schema` variable that's not trimmed, but I just want to double check first that there is a case where a FieldtypeMulti lacks a sort column, since I think they are all supposed to have one (again, I might be forgetting something).

pine3ree:
Hello @ryan,

I had a custom fieldtype (multi) representing apartment rates per year (and per seasons). Each year represents a table row and I need them ordered by year not by an extra/manual `sort` column.

Extending from Fieldtype does not cause any issue, but since it feels more natural extending FieldtypeMulti, I  added 
```php
//...
    public function getDatabaseSchema(Field $field)
    {
        $schema = parent::getDatabaseSchema($field);
        unset($schema['sort']); // Remove `sort` feature
//...
```
in order to drop the column.

`FieldtypeMulti::___savePageField` checks for the existence `sort` column, so I guessed that I could just remove it from my schema.
The problem arises when I load the field value, as by default orderby is usued with `"{$table.sort}"`. The solution I adopted is, overriding the `getLoadQuery` and removing the orderby clause

```php
    public function getLoadQuery(Field $field, DatabaseQuerySelect $query)
    {
        $query = parent::getLoadQuery($field, $query);
        $query->remove('orderby');
        return $query;
    }
```
kind regards

ryancramerdesign:
Thanks @pine3ree I understand now, I have added this.

matjazpotocnik:
@pine3ree can you close this PR?

PR #191, opened 2021-05-23
Use step attribute = precision in float inputfield if html5 number

pine3ree:
When using FieldtypeFloat its input fields should have a step attribute matching the field precision. This still allows any number of decimals, only change the step from the default "1" to the more related "precision" value. Since the decimal fieldtype already supported it, the code was moved to the float inputfield when rendering the html attributes and applied only when a positive precision is set.

### DIFF
--- a/wire/modules/Fieldtype/FieldtypeDecimal.module
+++ b/wire/modules/Fieldtype/FieldtypeDecimal.module
@@ -96,10 +96,7 @@ class FieldtypeDecimal extends Fieldtype {
 	public function getInputfield(Page $page, Field $field) {
 		/** @var InputfieldFloat $inputfield */
 		$inputfield = $this->wire()->modules->get('InputfieldFloat');
-		$precision = (int) $field->get('precision');
-		if($field->get('inputType') === 'number' && $precision > 0) {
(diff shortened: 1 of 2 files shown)


ryancramerdesign:
@pine3ree Thanks, I will add this. One question though: it looks like it adds the step attribute regardless of whether the inputType is 'number' or 'text'. Shouldn't the "step" attribute be omitted if the inputType is text? If you agree, don't worry about updating the PR as I can add that myself, but just wanted to double check in case the omission was intentional?

pine3ree:
Hello @ryancramerdesign ,

in getAttributesString the 1st line checks for it:
```php
if($attributes && $attributes['type'] === 'number') { 
```
so the step attribute is added only when the attribute 'type' is 'number'
kind regards

pine3ree:
Hello @ryancramerdesign ,

yes, correct, sorry, I had added that to my current project and forgot to add it back to my gh-fork. The purpose is obviously to add a default step based on precision, if not set programmatically (e.g. $myInputField->attr('step', '0.05').

ryancramerdesign:
Thanks @pine3ree I have added this.

matjazpotocnik:
@pine3ree can you close this PR?

PR #192, opened 2021-05-24
add skip-label configuration option for inputfield wrappers

pine3ree:
This option let us group/wrap different fields without annoying and/or redundant headers/labels

### DIFF
--- a/wire/core/InputfieldWrapper.php
+++ b/wire/core/InputfieldWrapper.php
@@ -1416,6 +1416,22 @@ class InputfieldWrapper extends Inputfield implements \Countable, \IteratorAggre
 			}
 		}
 
+		foreach($inputfields as $f) {
+			if($f->attr('name') != 'visibility') continue; 
+			/** @var InputfieldSelect $field */
+			$field = $this->modules->get("InputfieldSelect"); 
+			$field->attr('name', 'skipLabel'); 
+			$field->label = $this->_('Skip Label'); 
+			$field->icon = 'eye-slash';
+			$field->description = $this->_("Should the label for this fieldset be hidden in the editor?");
+			$field->addOption(self::skipLabelNo, $this->_('Never, show the fieldset name if empty'));
+			$field->addOption(self::skipLabelBlank, $this->_('Hide only if blank'));
+			$field->addOption(self::skipLabelHeader, $this->_('Yes, hide the header markup'));
+			$field->addOption(self::skipLabelMarkup, $this->_('Yes, hide all the header/label markup'));
+			$field->attr('value', (int) $this->skipLabel); 
+			$f->append($field); 
+		}
+
 		return $inputfields;
 	}
 


pine3ree:
I solved my simple requirement with a simple FieldtypeFieldsetOpen descendant, but I'll leave the p.r. open to see if it can be useful...maybe skipping the "skipLabel" option for FieldtypeFieldsetTabOpen (we want a label for the tab). Also note I did not include all the `Inputfield::skipLabel(...)` constants.

ryancramerdesign:
@pine3ree Thanks this looks like a good addition. I'm just wondering if this belongs in InputfieldFieldset rather than InputfieldWrapper? The InputfieldWrapper class is also used by anything wrapping Inputfields and includes non-fieldsets, such as InputfieldForm.

pine3ree:
Hello @ryancramerdesign ,
I added it to InputfieldWrapper because `InputfieldFieldsetOpen` extends it directly (see `FieldtypeFieldsetOpen`) and not InputfieldFieldset.

(I actually believe that it should not be available for `InputfieldFieldsetTabOpen` since we need a label in a tab)

Unrelated (maybe stupid) question (jsut to avoid you go looking at another "issue"):
Is there a reason for having field database tables being created for FieldtypeFieldset (open and close tab for instance)?

kind regards,
maks

ryancramerdesign:
@pine3ree I see, yes, InputfieldFieldsetOpen rather than InputfieldFieldset. But I suppose it's a similar question: should it go instead in InputfieldFieldsetOpen? (it's a little harder to find since it's an empty class in the FieldtypeFieldsetOpen.module file). Or maybe it goes in InputfieldFieldset but doesn't apply unless the class name includes `Fieldset` and lacks `Tab`? (this is probably the best route)

DB tables for fieldsets are just placeholders for consistency, in case anything operating on fields is doing anything with DB tables, it prevents the need additional conditional code. Maybe someday we'll use them for some kind of caching or search index, but for now they don't actually hold any data, and you should feel free to delete them if you find they are not needed in your installation.

pine3ree:
@ryancramerdesign , InputfieldFieldsetOpen (but not InputfieldFieldsetTabOpen) would be the most appropriate choice, unless other implementations of fieldset-like inputfield (extending directly from InputfieldWrapper) exist in user code or are planned for the core . As far as I know other then form and fieldset, only InputfieldMarkup extends from InputfieldWrapper.


##### PS: fieldtype-fieldset db-tables:
Should we override `__createField` for no-op?
Having a db-table for FieldtypeFieldset fields with actual data in it would mean having at least a row of data for each page using that fieldset. But fieldsets do no contain data, they are only used for presentation, which is the same for all instances. If we insert page's field specific data into the db for a fieldset, that data should be editable per page-field and we would need other fields for that. But then those other fields would have their own database tables....
The one limit case where data could be stored for a fieldset, is when we want to store all the inner fields data in one concat field for simpler search selector purpose, but searchable fields usually do not belong to the same fieldset and I use to create an ad-hoc field for that populated via beforeSave, saveReady hooks.

PR #194, opened 2021-05-30
Add base style via PHP instead of LESS @import

BernhardBaumrock:
While creating https://github.com/baumrock/AdminStyleRock I realized that it is better to add the base style via PHP instead of using an @import statement in the descending less styles.

On the one hand this means that it does not matter where the style lives the import will always find the correct core base style and on the other hand the need for the initial @import makes the style unnecessarily complicated because you need to know where the style lives and you need to be sure to set the correct path. That's something you should not need to think of when building a new style!

This PR means that you can create an empty custom style and it will work (showing the base style). Without that commit you'd get a broken backend style because it does not include the base style.


matjazpotocnik:
@BernhardBaumrock can you close this?

BernhardBaumrock:
sure, thx :)

PR #195, opened 2021-06-05
remove deprecated method calls (FilenameArray::unique)

pine3ree:


matjazpotocnik:
And perhaps remove code that targets IE7?
```
	<!--[if lt IE 8]>
	<link rel="stylesheet" type="text/css" href="<?php echo $config->urls->adminTemplates; ?>styles/ie7.css" />
	<![endif]-->
```

PR #204, opened 2021-12-06
Add links to quickly edit a file in other languages

MrSnoozles:
When editing translations, this adds links to the other language versions of the file.

**Before:**
grafik

**After:**
grafik

Without these links, the workflow to edit a translation in another language is **very cumbersome** and takes 4 clicks plus some searching:
* click on Languages
* choose your language
* choose the file
* click on edit


ryancramerdesign:
Thanks @MrSnoozles I have added this PR. I made a few adjustments, mostly to accommodate the case of when a translation file does not yet exist for a particular language you click on, it now creates it rather than showing a fatal error message.

PR #210, opened 2022-01-04
add $files->url() and $files->path()

BernhardBaumrock:
hey @ryancramerdesign 

What do you think of this addition to WireFileTools? It's an old request from 2019 and I keep stumbling over it again and again. Lately jens had a similar need that would also be obsolete if we had `$files->url() and $files->path()`

My original request used `$config->url() and $config->path()` syntax but I had a look at the config class and it already has such methods. That's why I added the feature to the WireFileTools which have no similar methods and should therefore not add any problems or confusion!

This is what I used for testing:

```php
echo "### url() ###";
db($files->url("/var/www/html/site/assets/cache/TracyDebugger/consoleCode.php"));
db($files->url("/foo/bar"));
db($files->url("/site/assets/foo.jpg"));
db($files->url("assets/foo.jpg"));
db($files->url($config->paths($modules->get('TracyDebugger'))));
db($files->url($config->paths->root));

echo "### path() ###";
db($files->path("/foo/bar.jpg"));
db($files->path($config->urls($modules->get('TracyDebugger'))));
```

Results:

```
### url() ###
'/site/assets/cache/TracyDebugger/consoleCode.php'
'/foo/bar'
'/site/assets/foo.jpg'
'/assets/foo.jpg'
'/site/modules/TracyDebugger/'
'/'
### path() ###
'/var/www/html/foo/bar.jpg'
'/var/www/html/site/modules/TracyDebugger/'
```

### DIFF
--- a/wire/core/WireFileTools.php
+++ b/wire/core/WireFileTools.php
@@ -61,6 +61,47 @@ class WireFileTools extends Wire {
 	}
 
 	/**
+	 * Given a filepath/url of a file/folder return the url relative to pw root
+	 * 
+	 * Note that this method does not check wether the file exists or not!
+	 * It uses simple string comparisons.
+	 * 
+	 * Usage:
+	 * 
+	 * $files->url("/var/www/html/site/assets/demo.jpg");
+	 * --> /site/assets/demo.jpg
+	 * $files->url("/site/assets/demo.jpg"); // with leading slash
+	 * --> /site/assets/demo.jpg
+	 * $files->url("site/assets/demo.jpg"); // no leading slash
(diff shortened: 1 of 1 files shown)


BernhardBaumrock:
This is the mentioned issue: https://github.com/processwire/processwire-requests/issues/326

elabx:
Hi! Any reason this was closed? Is there an alternative? Thanks!

BernhardBaumrock:
I don't know @elabx

BernhardBaumrock:
@elabx just got feedback by Ryan that he is not willing to add this. At least we finally know.

PR #214, opened 2022-01-19
Make previewColumns configurable for Fieldtype / Inputfield Selector

adrianbj:
It's really useful to be able to define the columns shown when clicking the "show" links for viewing results. This supports that like this:
image

I actually think it would be nice to add other options too, like how edit / view links should be handled etc, but wanted to see if you'd at least be willing to support the changed in this PR first.

### DIFF
--- a/wire/modules/Fieldtype/FieldtypeSelector.module
+++ b/wire/modules/Fieldtype/FieldtypeSelector.module
@@ -6,7 +6,7 @@
  * Concept by Antti Peisa
  * Code by Ryan Cramer
  * Sponsored by Avoine
- * 
+ *
  * ProcessWire 3.x, Copyright 2016 by Ryan Cramer
  * https://processwire.com
  *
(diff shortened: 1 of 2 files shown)


adrianbj:
PS, sorry about the whitespace changes - my editor cleans up spaces at the end of lines automatically, as it should :)

BernhardBaumrock:
> PS, sorry about the whitespace changes - my editor cleans up spaces at the end of lines automatically, as it should :)

I think you are using VSCode. I've rearranged my settings.json so that all tidy options are at one place, then I can easily comment them out before making changes to the core:

img

adrianbj:
Thanks @BernhardBaumrock - I probably should do that, but PW is the only OS code I've contributed to that doesn't automatically do this, and I've never understood why. Still, it's not for me to judge, so I will try to be better next time :)

BernhardBaumrock:
Sure, you are right that it would be better if PW did that already, but as long as it doesn't I think we should keep files as they are so the diff is better readable :)

@ryan it seems to be very simple to fix that: https://stackoverflow.com/a/57629631/6370411

I could also do that and add a PR, but from a security perspective I think this should be done by you, as a PR with thousands of changed files might not really be reviewable...

Here is the link to PhpStorms reformat feature (looks also easy): https://www.jetbrains.com/help/phpstorm/reformat-and-rearrange-code.html#exclude_part_of_code

teppokoivula:
Just a quick note from the sidelines: perhaps discussion about code formatting should be a new feature request? I'd argue that there are reasons for and against, and comments of mostly unrelated PR may not be the best place to discuss those :)

adrianbj:
Thanks Ryan!

PR #216, opened 2022-02-03
add form addAfter() and addBefore()

BernhardBaumrock:
Without the PR:

```php
$f = $this->wire(new InputfieldMarkup());
$f->value = 'foo bar';
$f->label = 'my field';
$form->insertAfter($f, $form->get('title'));
```

With this PR:

```php
$form->addAfter('title', [
  'type' => 'markup',
  'label' => 'my field',
  'value' => 'foo bar',
]);
```

### DIFF
--- a/wire/core/InputfieldWrapper.php
+++ b/wire/core/InputfieldWrapper.php
@@ -275,6 +275,52 @@ class InputfieldWrapper extends Inputfield implements \Countable, \IteratorAggre
 		}
 		return $this; 
 	}
+	
+	/**
+	 * Add a new item after an existing item
+	 * 
+	 * Usage:
+	 * $form->addAfter('title', [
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
@BernhardBaumrock Thanks I have added this, though in a different way. I updated the existing insertBefore() and insertAfter() methods to also support what you are doing with addBefore() and addAfter(). Very good ideas with those methods you added, but I was confused by the argument order of addBefore() and addAfter() since it was different from the order stated in the method name. Plus I thought it was just simpler to improve the existing methods that already do this by expanding what their arguments accept, rather than adding new similar methods for this purpose.  This is a very useful addition, thank you for suggesting it.

BernhardBaumrock:
Great! Thx @ryancramerdesign :)

PR #219, opened 2022-03-08
add button to close all notifications at once

BernhardBaumrock:
Hey Ryan, this is an old issue dating back to 2014: https://processwire.com/talk/topic/8329-how-to-closedelete-notifications/

I always get lots of - persistent - notifications after login:

img

I need to close every single one of them. This PR adds an additional X to close all at once:

image

Note that I've only updated the unminified JS

### DIFF
--- a/wire/modules/System/SystemNotifications/Notifications.css
+++ b/wire/modules/System/SystemNotifications/Notifications.css
@@ -109,6 +109,11 @@
 	font-family: Arial, sans-serif;
 	font-size: 14px; 
 }
+.NotificationBug .closeAll {
+	padding-left: 10px;
+	padding-right: 5px;
+	font-weight: normal;
+}
 
 .NotificationBug.open,
 .NotificationBug:hover {
--- a/wire/modules/System/SystemNotifications/Notifications.css
+++ b/wire/modules/System/SystemNotifications/Notifications.css
@@ -108,9 +108,10 @@
 .NotificationBug .qty {
 	font-family: Arial, sans-serif;
 	font-size: 14px; 
+	padding-right: 5px;
 }
 .NotificationBug .closeAll {
-	padding-left: 10px;
+	padding-left: 5px;
 	padding-right: 5px;
 	font-weight: normal;
 }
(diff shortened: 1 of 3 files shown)


BernhardBaumrock:
Hey @ryancramerdesign I get now what you were talking about:

img

There seems to be different logics for showing notifications in different places. I'm talking about the notifications that pop up when the systemNotifications module is installed after logging in or after refreshing modules.

The screenshot above is taken after a page was saved. There it is possible to close all notifications at once.

BernhardBaumrock:
turns out that there are some inconsistencies between the regular notifications and the ones coming from SystemNotifications module: https://github.com/processwire/processwire-issues/issues/1554

ryancramerdesign:
@BernhardBaumrock I see, I didn't catch you were using the SystemNotifications module at first. This is a pretty old module that I've been thinking we'll remove from the core, as I don't think it really belongs in the core anymore. I thought it might evolve better in its own repo, as I think it's still a good module, but too much in the core, as notifications are now the purview of the admin themes.  A lot of what it does was built to the needs of the company that sponsored it, but for most I think it's better to leave presentation of notifications to the admin theme. At this time the SystemNotifications module was written, we didn't have an AdminTheme framework like we do now.

BernhardBaumrock:
Thx @ryancramerdesign that is an interesting information! The reason why I have installed this module is to get at least some kind of warning if another user is editing the same page at the same time. Sounds like this is not a good option any more.

I prefer the default look and behaviour of notifications anyhow, so I'd be happy to remove the module. The only thing that I really need is the notification about same page editing. Maybe this feature could be built into the core as a lightweight alternative to the pro activity module?

ryancramerdesign:
@BernhardBaumrock Do you have the UserActivity module? This is the best one for preventing page edit collisions, and it also highlights pages being edited before you even click to edit them. It's a really useful module, I might have to add it to the core.

BernhardBaumrock:
Hey Ryan, seems I forgot that I have bought ProDevTools :D That means I'll install the UserActivity module instead of the SystemNotifications. I'll close this PR as the SystemNotifications module seems obsolete...

Having UserActivity in the core would be awesome!! I think this is something that every site could really benefit from. That's always one of the CONS I have to admit when showing ProcessWire to someone: "Oh, yes, you need to be careful not to overwrite changes that someone else did at the same time... ProcessWire does not warn you in such situations..."

PR #220, opened 2022-04-05
Update InputfieldPassword.module

mpsn:
ComplexifyFactor does not work with international locales with decimal comma (e.g. German) below PHP 8. Behaviour changed in PHP since PHP 8: https://wiki.php.net/rfc/locale_independent_float_to_string


matjazpotocnik:
@mpsn can you verify that [this fix ](https://github.com/processwire/processwire/commit/2a8899dcd977fc413dafbd33c7edc494c8ebbfab) also works for you?

mpsn:
@matjazpotocnik Thanks for pointing me to this fix and yes, it also works for me.

PR #221, opened 2022-05-03
Add option to specify default page to load images from/to

donatasben:
Useful if you want to keep all user images in one place (page) by default.
Initially the dialog will show the selected page images by default - admin editors will not have to choose a page manually.
Code impact: minimal.
Adding one additional configuration option to the "ProcessPageEditImageSelect" module.


donatasben:
Hello,
I suggest a very small addition to "ProcessPageEditImageSelect" module ("pwimage" CKEditor plugin).
The new offered module configuration option allows to select a default page to load images from/to when insert an image into CKEditor enabled field.

Benefits:

- Easy to set up a single place for all images that are used in rich text fields.
- Content editors don't have to remember to choose a different source page each time they upload an image to rich text field.

wire/modules/Process/ProcessPageEditImageSelect/ProcessPageEditImageSelect.module
```diff
@@ -289,6 +290,11 @@ class ProcessPageEditImageSelect extends Process implements ConfigurableModule {
 		}
 		
 		if(!$id) throw new WireException("No page specified");
+		if($this->sourcePage) {
+			$this->page = $this->pages->get($this->sourcePage);
```
teppokoivula, line 294: Not particularly familiar with how this module works, but should this be findOne instead of get? Just wondering what would happen if current user doesn't have access to the page, since get bypasses access/status checks :)

PR #223, opened 2022-05-06
WireMail : Use 'fromName' from config.

hiboudev:
In site/config.php I added :

> $config->wireMail = array(
> 	'module' => '', 
> 	'from' => 'myname@example.com', 
> 	'fromName' => 'My Name', 
> 	'headers' => array(), 
> 	'blacklist' => array()
> );

But fromName value is not used in WireMail.php.
It looks that it's the same for headers and blacklist but I just added fromName actually.

My use case is when using LoginRegister, fromName is not used, but with ProcessForgotPassword it's used, so I'm not sure my fix is at the good place but it works for my case.

### DIFF
--- a/wire/core/WireMail.php
+++ b/wire/core/WireMail.php
@@ -636,7 +636,11 @@ class WireMail extends WireData implements WireMailInterface {
 		if(!strlen($from)) $from = $this->wire('config')->adminEmail;
 		if(!strlen($from)) $from = 'processwire@' . $this->wire('config')->httpHost;
 		
-		$header = "From: " . ($this->fromName ? $this->bundleEmailAndName($from, $this->fromName) : $from);
+		$fromName = $this->fromName;
+		
+		if(!strlen($fromName) && !empty($settings['fromName'])) $fromName = $settings['fromName'];
+		
+		$header = "From: " . ($fromName ? $this->bundleEmailAndName($from, $fromName) : $from);
 
 		foreach($this->header as $key => $value) {
 			$header .= "\r\n$key: $value";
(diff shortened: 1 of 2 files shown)


ryancramerdesign:
Thanks @hiboudev, good idea, I have added your PR

hiboudev:
@ryancramerdesign How does pull request work? I tried to add another one but I'm back on this old one. Shouldn't you close it?

The last one is a bug fix on page reference field.

I have a page reference field with a query string = "title carClasses.title", it targets the "car" template that has a carClasses field that is another page reference with selector string = "game.gameId=page.game.gameId, sort=title".

This page reference fields use Autocomplete with multiple pages.

And it's not working cause in FieldtypePage.getMatchQueryCustom(), line 1171, the regex is stopping before the second dot when removing selector. It should remove "game.gameId=page.game.gameId" but it removes only "game.gameId=page.game" and ".gameId" is still there.

Fixed regex is:
`/[a-zA-Z0-9_.]+\s*[=!]+\s*page(\.[a-zA-Z0-9_]+)+\s*,?/`

instead of:
`/[a-zA-Z0-9_.]+\s*[=!]+\s*page\.[a-zA-Z0-9_]+\s*,?/`

hiboudev:
There's something I don't understand, why do we need to use the selector on the second page field?

Page1->field1 is a page reference to Page2 type, field1 also uses a selector "field2=xxx". Page2->field2 is also a page reference field with a custom selector, but why do we need to process it?

I am testing, and if I remove this whole code, the issue is fixed, and without CPU usage issue, selector "game.gameId=page.game.gameId" is working:

```
//		if($findPagesSelector) {
//			// remove the existing include=hidden|all|unpublished, only if selector specifies a different include=something
//			if(strpos($findPagesSelector, 'include=') !== false) {
//				$selector = preg_replace('/include=[a-z]+,/', '', $selector);
//			}
//			$selector .= $findPagesSelector . ", ";
//			if(strpos($selector, 'page.') !== false) {
//				// remove any page.property from selector as it won't be applicable during a getMatchQuery
//				$selector = preg_replace('/\bpage\.[a-zA-Z0-9_]+\s*[=!]+\s*[^,]*,?/', '', $selector);
//				$selector = preg_replace('/[a-zA-Z0-9_.]+\s*[=!]+\s*page(\.[a-zA-Z0-9_]+)+\s*,?/', '', $selector);
//			}
//		}
```

PR #224, opened 2022-05-07
Add option: Disable Link Text Edit

FlipZoomMedia:
The feature introduced in PR#38, that the link text can be edited in the modal window, has the problem that if there is markup in the link text, such as icons, the markup is lost when changing it. For this the option was implemented to deactivate this function from PR#38 (https://github.com/processwire/processwire/pull/38).


ryancramerdesign:
Thanks @FlipZoomMedia I have added this.

PR #225, opened 2022-05-19
Fix PHP 7.x warning: count()

FlipZoomMedia:
In various places count() is still used. This results in warning messages in PHP 7.2.x and higher: "Warning: count(): Parameter must be an array or an object that implements Countable". I have replaced all of them with wireCount().

### DIFF
--- a/wire/modules/Process/ProcessPageEditImageSelect/ProcessPageEditImageSelect.module
+++ b/wire/modules/Process/ProcessPageEditImageSelect/ProcessPageEditImageSelect.module
@@ -27,7 +27,7 @@ class ProcessPageEditImageSelect extends Process implements ConfigurableModule {
 		return array(
 			'title' => 'Page Edit Image',
 			'summary' => 'Provides image manipulation functions for image fields and rich text editors.',
-			'version' => 120,
+			'version' => 121,
 			'permanent' => true,
 			'permission' => 'page-edit',
 		);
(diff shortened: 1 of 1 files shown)


BernhardBaumrock:
@FlipZoomMedia I've never seen any of those warnings... where exactly do you get them? Are you sure the problem is with the core rather than a custom implementation on your end where you throw something into count() that is not a proper array or object?

FlipZoomMedia:
Hi @BernhardBaumrock I checked again and indeed this phenomenon only occurs within repeaters. The rest from the system is base and a fresh PW 3.0.200 master, current RepeaterMatrix 0.0.8 on PHP 7.4.X with PW debug mode on. See screenshots.

A simple repeater item is placed and an inline image is inserted inside the CKE. Automatically ProcessPageEditImageSelect opens the path of the repeater page, there the Count() warning appears. Maybe some more hand should be done here, that the user lands on the parent page of the repeater item as path and not in the repeater page itself?

screenshot-STAGE-2022 05 20-14_58_45

screenshot-STAGE-2022 05 20-14_58_44

matjazpotocnik:
@FlipZoomMedia I can't replicate on PHP 8.1.6 and PW 3.0.200. $images are always an array, so count($images) works. I don't have RepeaterMatrix to test.

BernhardBaumrock:
Hi @FlipZoomMedia thx for that details. I think that will help ryan to identify the problem.
It just didn't feel good to me to hide the problem by using wireCount as in my opinion everything should work when using count (which will be the case if only arrays or countable objects make there way into the function call).

ryancramerdesign:
@FlipZoomMedia Thanks I have added this. 

@BernhardBaumrock While I'm not seeing this particular issue, some of the cases the PR corrects are those where the value to count() is expected to be an array or null. So in this case replacing count() with wireCount() is I think the right solution. The code was written when null was an expected and valid argument for count(). PHP later changed the purpose and meaning of count(), which is why we added wireCount() to maintain the original purpose, and I still find the original behavior quite useful.

PR #226, opened 2022-05-19
Fix assignment-in-conditional bugs

netcarver:
Fixes two incorrect assignments in conditional tests during fieldgroup saves.


matjazpotocnik:
@netcarver, Steve can you close this PR?

PR #227, opened 2022-06-17
Partial fix for dnd image replacement issue #1586

pine3ree:
Seems to solve one of the issues (issue3 ) described here: https://github.com/processwire/processwire-issues/issues/1586
In short: when using a MULTIPLE-image field with custom fields, editing one image to replace just the image after drag and drop custom fields values are lost (see last comment on the issue for more details). Replacing it again wiyhout saving the page will still make you think the image description is lost, but if you save the page you'll end up with 2 image files.


pine3ree:
Closing it as it was included with mods https://github.com/processwire/processwire/commit/c697be7ddd3806541694688768170cec78b01b12 by @ryancramerdesign

PR #228, opened 2022-06-20
exit for-loop at first match

pine3ree:


ryancramerdesign:
Thanks @pine3ree I have added this

PR #229, opened 2022-06-24
possible fix for issue #1586  (sub-issue 1)

pine3ree:
Hello @ryancramerdesign , I haven't tried all the combinations but adding a same name check is solving my issue. Without the check the code removes the file with the same name from the server thinking that I want to replace it with other filename, but then it tries to rename the deleted file. I haven't had the time to read all the related code, but maybe having identified the line in the code could be a useful hint. Are you still unable to reproduce (sub)issue 1?   

kind regards

related comment: https://github.com/processwire/processwire-issues/issues/1586#issuecomment-1165722370


ryancramerdesign:
@pine3ree While I haven't been able to observe the the issue that this fixes, the suggested logic makes sense to me, and I trust you that it needed this fix, so I have added. Thanks.

PR #230, opened 2022-06-25
fix "typo" in _init.php

pine3ree:


ryancramerdesign:
Good catch, I've added this, thanks.

PR #231, opened 2022-06-26
simpler arrayToCSV builder

pine3ree:


ryancramerdesign:
Thanks @pine3ree that's much simpler, I have added

PR #232, opened 2022-06-27
fix pagination issues for uncentered current page and unneeded separator 

pine3ree:


matjazpotocnik:
@pine3ree Maks, could you check this thread about pagination: https://processwire.com/talk/topic/21912-change-how-many-page-numbers-to-show-between-ellipses-in-pager/ and see if your version works as it should? There is also my version of PagerNav.php attached later in the comment. I'm on vacation now with limited internet access and can't test.

pine3ree:
Hello, @matjazpotocnik...I am a little busy on a jub right now, but I will read it as soon as I can. I am not sure that the uncentered current page is an issue or just a feature...but I believe that the hellip separator between consecutive page-links is. I just opened this PR, just as reference for a possible solution if RC wants to take a look at it. I am using my own class (which has first/last unnumbered links on the outside, as in the pictures in the issue). I just tried to quickly blend the pw PagerNav to what I am used. :-) Apart from the number of visible page-links which I usually set  to 9 (4 on each side of current), I the use page-items (list items) hidden on mobile using css classes. best regards and...have a good time!!

ryancramerdesign:
@pine3ree Thanks, looks like good improvements. I'm reluctant to pull this in because I have so many projects that depend on the current pagination behavior, and it's been so widely tested in so many places. If we could reduce it to just fixing anything that's broken that would be preferable. Or if this is more specific to new pagination logic, would you consider releasing it as a module? The core MarkupPagerNav is just a module, so it is easily swappable.

PR #235, opened 2022-07-20
use uppercase sql keyword

pine3ree:


ryancramerdesign:
@pine3ree As far as I know, MySQL isn't case sensitive with this?

adrianbj:
I think it's more about following convention and the improved readability that uppercasing gives.

pine3ree:
> @pine3ree As far as I know, MySQL isn't case sensitive with this?

Hello, @ryancramerdesign. It is not case-sensitive indeed, but all other SQL reserved words are UPPERCASED elsewhere. So it's more like a convention used in most of, if not all, the db-layer libraries I use. kind regards.

PR #237, opened 2022-07-26
simplify row number calc

pine3ree:
Hello @ryancramerdesign ,

I was originally digging into this part of pw-code to ask you a question:
Is there a way to trigger/force the usage of a textarea field?

Most of the times I prefer using language placeholders like the following:
```php
__('privacy_intro');
__('label_email');
__('label_full_name');
// etc..
```
instead of the full english/default language text. I find it easier to identify and to manage translatable strings. But then, w/o newlines or w/o long text a textarea inputfield is never created.

Can we add a further check like "if the text ends in `_textarea` or `__textarea`" like 
```php
__('privacy_intro_textarea');
__('privacy_intro__textarea');
```
then a InputfieldTextarea is used?

kind regards,
maks

### DIFF
--- a/wire/modules/LanguageSupport/ProcessLanguageTranslator.module
+++ b/wire/modules/LanguageSupport/ProcessLanguageTranslator.module
@@ -380,11 +380,10 @@ class ProcessLanguageTranslator extends Process {
 		/** @var InputfieldText|InputfieldTextarea $field */
 
 		if(strpos($untranslated, "\n") !== false) {
-			$qty = substr_count($untranslated, "\n")+1;
+			$qty1 = substr_count($untranslated, "\n")+1;
 			$qty2 = substr_count($translated, "\n")+1; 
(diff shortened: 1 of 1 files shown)


pine3ree:
PS
We could also add the number of rows: `__textarea_3`

ryancramerdesign:
@pine3ree The translation functions are meant to contain translatable text rather than tokens or variable names. But I also see no harm with using them how you are, other than that a translation is always required, and thus wouldn't work in a single-language environment. Because of that, I don't think that adding things like `_textarea` to the translatable text is a good way to go, but I like the idea of being able to specify these things along with it somehow. I've added something similar to your suggestion, which instead can take the Inputfield type and rows values from the translation comment, i.e.
~~~~~
// rows greater than 1 always assumes textarea
__('text to translate'); // rows=3

// or you can specify input type
__('text to translate'); // type=textarea

// or both
__('text to translate'); // type=textarea rows=10
~~~~~

PR #238, opened 2022-07-28
allow custom precision float to be returned from Pageimage::ratio()

pine3ree:


ryancramerdesign:
Thanks @pine3ree this has been added

PR #239, opened 2022-07-30
fix php-doc block typo

pine3ree:


ryancramerdesign:
Thanks @pine3ree good find, I've added it

PR #240, opened 2022-08-18
fix return type in php-doc block

pine3ree:
Should we also add `string` for the `$languageID` param both in setter and getter, like in the `LanguagesPageFieldValue` implementation?


ryancramerdesign:
Thanks @pine3ree I have added this.

PR #241, opened 2022-08-25
Handle concatenated value for FieldtypeOptions

dadish:
Update `wakeupValue` method of the `FieldtypeOptions` module to handle a case when the value is a concatenated string containing multiple values.

Should fix the [#1616](https://github.com/processwire/processwire-issues/issues/1616)


ryancramerdesign:
Thanks @dadish I have added this

dadish:
d02efe55f2ed3cc6964c93db54fcaeda2a6baac8

PR #242, opened 2022-09-09
Update ProcessPagesExportImport.module

jploch:
Support for exporting/importing $page->meta() data via core ProcessPagesExportImport Module. Tested with multiple pages and works well so far.

I made changes to the PagesExportImport Module, but it might be better to implement this in the core [PagesExportImport Class](https://processwire.com/api/ref/pages-export-import/). That would make the feature available to the API and other modules as well. But I am not sure how to do that, because to import the meta data, the pages must be created first, so the code might only work in the module, after the pages are created.


jploch:
Sorry for not signing the CLA license. I just did it now. So maybe you can have another look. Would be nice to have this in the core.

PR #243, opened 2022-09-15
Update PHPDoc in WireArray

MrSnoozles:
When calling `return PageArray([$page])->unique();` I get a warning from the IDE that my method expects a PageArray, but a WireArray is returned.  This is because the documentation for some methods in WireArray does not consider derived classes.

### DIFF
--- a/wire/core/WireArray.php
+++ b/wire/core/WireArray.php
@@ -203,7 +203,7 @@ class WireArray extends Wire implements \IteratorAggregate, \ArrayAccess, \Count
 	 * 
 	 * #pw-internal
 	 *
-	 * @return WireArray
+	 * @return $this
 	 *
 	 */
 	public function makeNew() {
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
@MrSnoozles Thanks for the PR, what point you make is a good one. The thing is that, unless something has changed, `$this` refers to the current instance, meaning the exact same instance of the class that it was called from. This indicates it can be used for a fluent interface. 

Most of the changed methods return a new instance, which is different from `$this`. So when it says `@return WireArray` it seems more technically correct, even if not as convenient. 

I wonder if `@return self` or `@return static` might accomplish it better? I'm not totally clear on the differences, as it seems they were interchangeable with $this awhile back, or maybe I just didn't understand. But I think that `self` may be intended to refer to an instance of the current class (whether same instance or another)? I'm trying to search and find the answer but getting conflicting info. But it `self` means what I think it does, that would probably be the right one to use for this case.

PR #244, opened 2022-09-16
New option: Remember Language Tabs Focus

FlipZoomMedia:
This new option gives editors a better workflow when working with languages and the Language Tab Module. 

screenshot-localhost-2022 09 16-15_12_36

So far, the focus of the current language was lost after each save and the editor always had to switch back to the language they were currently editing. 

https://user-images.githubusercontent.com/708183/190647573-d7f59f93-c277-4fa7-a9e9-60b96ecb2d8e.mp4

When editing a language version for a longer period of time, the new option now improves the workflow. By double-clicking on the desired language version, the status is saved in a cookie for the current session if the Remember option is activated. After each save, the focus is now automatically set to the last selected language. 

https://user-images.githubusercontent.com/708183/190648003-547fdf77-ea93-40f2-9e65-91fafd5d4d3c.mp4

See original ticket: [processwire/processwire-issues#1622](https://github.com/processwire/processwire-issues/issues/1622)

### DIFF
--- a/wire/modules/LanguageSupport/LanguageTabs.js
+++ b/wire/modules/LanguageSupport/LanguageTabs.js
@@ -24,6 +24,7 @@ function dblclickLanguageTab(e) {
 	setTimeout(function() {
 		clickLanguageTabActive = false;
 	}, 250);
+	if(ProcessWire.config.LanguageTabs.rememberDbClickedTabs === 1) setLanguageTabsCookie(langID);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


ryancramerdesign:
@FlipZoomMedia Somehow I missed that we were talking about DOUBLE click before. I thought we were just talking about clicking once on a language tab in a field. Given that double-click action, I don't think it is necessary for it to be a configurable setting... I doubt anyone would have an issue with it remembering that. As a result, I've added your PR but simplified it a bit, removing the configuration setting and just making it always enabled.

FlipZoomMedia:
Thanks @ryancramerdesign, The username is correct, but my GitHub mail is david.karich@flipzoom.de - maybe that's the problem. 🤷🏻‍♂️

FlipZoomMedia:
I had not added the info@ to my profile until now, GitHub seems to reference only by mail address and not usernames. Added the info address, now it displays and links correctly. Thanks a lot Ryan. I will test your implementation on Monday.

FlipZoomMedia:
Works perfectly. Thank you very much.

PR #246, opened 2022-10-18
Add method to read config from JSON file

BernhardBaumrock:
Hey @ryancramerdesign I have a setup where I need to write the config of the site dynamically via PHP. Manipulating the original config.php file is really NOT easy... But it would be trivial to write the new config as JSON file. For that to work it would be nice to have this:

```php
$config->readFromFile('/path/to/config.json');
```

### DIFF
--- a/wire/core/Config.php
+++ b/wire/core/Config.php
@@ -780,6 +780,22 @@ class Config extends WireData {
 	}
 
 	/**
+	 * Read config from json file
+	 *
+	 * Usage: Add this line in your /site/config.php
+	 * $config->readFromFile('/path/to/your/config.json');
+	 *
+	 * @param string $file
+	 * @return void
+	 */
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
@BernhardBaumrock Sounds good to populate from a JSON file, but just wondering if we need a dedicated method for it (?), as it's already pretty simple:
~~~~~
$config->data(json_decode(file_get_contents('file.json'), true));
~~~~~

BernhardBaumrock:
Hey @ryancramerdesign thx for having a look! Thx for $config->data() - I've never used that. Nevertheless I'd prefer a dedicated method for two reasons:

1) I think my version is easier to read/understand
2) My version will end up in the api docs which will make it visible to others that might not have the idea that what you wrote is even possible

But it's not a big thing, you are right :)

PR #249, opened 2022-11-06
Fix minor typo in ProcessCommentsManager.module

v-maillard:
Changed "meata" to "meta" ([ProcessCommentsManager.module#L394](https://github.com/processwire/processwire/blob/dev/wire/modules/Process/ProcessCommentsManager/ProcessCommentsManager.module#L394))


ryancramerdesign:
Thanks @v-maillard it looks like this one was already fixed at some point though not sure when. But I'll mark this as completed.

PR #251, opened 2022-12-11
Add download button to InputfieldImage.

JanRomero:
Hi, the idea for this came from [this forum thread](https://processwire.com/talk/topic/27709-who-can-help-with-adding-a-download-icon-to-image-fields/) by @BernhardBaumrock. It’s a small but imho quite useful addition.

I’ve tried to keep changes to a minimum. It’s virtually entirely a CSS thing because the download button is a simple <a> element, but to make it work nicely I had to make the CSS independent of the label element and I renamed .gridImage__trash to .gridImage__btn to make it more generic. This somewhat opens the door for more such buttons – it actually clutters up the UI surprisingly little: 
grafik
(Obviously that’s just a dumb mockup. The actual PR only adds the second button, [as pictured here](https://processwire.com/talk/topic/27709-who-can-help-with-adding-a-download-icon-to-image-fields/?do=findComment&comment=227744).)

Buttons will be in reverse order in “list” view.

I’ve tested it successfully with both stock themes. Hope it’s not too messy. Thanks for your consideration!

### DIFF
--- a/wire/core/ProcessWire.php
+++ b/wire/core/ProcessWire.php
@@ -80,7 +80,7 @@ class ProcessWire extends Wire {
 	 * Reversion revision number
 	 * 
 	 */
-	const versionRevision = 239;
+	const versionRevision = 240;
 
 	/**
 	 * Version suffix string (when applicable)


BernhardBaumrock:
Hey @JanRomero just read a similar request here: https://github.com/processwire/processwire-requests/issues/392#issuecomment-1367795930

Would it be possible to adopt your PR so that it also accounts for file fields?

JanRomero:
Hey @BernhardBaumrock, this PR only touches files specific to InputfieldImage, so the changes don’t generalise to InputfieldFile, but it’s very simple to do for InputfieldFile since there is only one UI mode and the delete button is already floated:

```PHP
$this->addHookAfter('InputfieldFile::renderItem', function(HookEvent $event) {
    $event->return = str_replace('</p>',
    <<<HTML
        <label class='InputfieldFileDelete'>
            <a href='{$event->arguments(0)->url}' download>
                <i class='fa fa-fw fa-download'></i>
            <a>
        </label></p>
    HTML, $event->return);
});
```
Added download button

That’s without any changes to the styling. A little margin might be nice, and the CSS class should be renamed, I guess.

Alternatively it’s even easier to just make the link that already exists download by default, while still allowing “open in new tab” using the usual browser features:

```PHP
$this->addHookAfter('InputfieldFile::renderItem', function(HookEvent $event) {
    $event->return = str_replace("<a class='InputfieldFileName'", "<a class='InputfieldFileName' download", $event->return);
});
```

Not sure Image and File should be combined in the same PR?

JanRomero:
@ryancramerdesign Hi, thanks for the changes and the heads up on the other PR :) I submitted the CLA with my Github user info and fudged the commit to be from me instead of “Anonymous”.

ryancramerdesign:
@JanRomero Thanks, I have partially added this PR. I'm not convinced we need a dedicated download button, but really liked the way you approached the addition of it. So I took your update and turned it into a broader update that adds new hookable functions where you can add whatever icon-buttons you want to the image, and then did the same with the buttons that appear in the image edit mode. I put "Download file" examples in the phpdoc for these new methods so people can see how to easily add their own download buttons, or any other kind of buttons.

PR #252, opened 2022-12-12
MarkupRSS: Allow markup in <description>

JanRomero:
Me again ✌️. I want my RSS feed items to have the full content, not just a summary. [According to the spec](https://www.rssboard.org/rss-specification#hrelementsOfLtitemgt), the `<description>` field is supposed to be used for this, but MarkupRSS does not allow it for two reasons:

1. It doesn’t convert relative to absolute links.
2. It double-encodes markup by enclosing entity-encoded content in CDATA. This makes (most?) feed readers show the markup instead of rendering it. Both entity escaping and CDATA should be rendered as per the spec, but the combination of both doesn’t work.

This PR removes entity encoding from the `<description>` element and adds the URL conversion that was previously only applied to `<content>`. It also extends this conversion to account for single quotes such as `href='/path'`.

Here is another link on supplying only the `<description>` element:
https://validator.w3.org/feed/docs/warning/NeedDescriptionBeforeContent.html

Thank you!

### DIFF
--- a/wire/modules/Markup/MarkupRSS.module
+++ b/wire/modules/Markup/MarkupRSS.module
@@ -222,13 +222,23 @@ class MarkupRSS extends WireData implements Module, ConfigurableModule {
 			}
 		}
 		
+		$rootUrl = $this->wire()->config->urls->httpRoot;
+		
 		if($this->itemDescriptionField) {
 			// description summary
 			$description = $page->get($this->itemDescriptionField);
 			if($description !== null) {
 				$description = $sanitizer->unentities($description, true);
 				$description = $this->truncateDescription($description);
-				$description = '<![CDATA[' . $this->ent($description) . ']]>';
+				if(strpos($description, '/') !== false) {
+					// convert relative URLs to absolute with host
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
@JanRomero Thanks, I have added something similar to what you've suggested here. The intention with the module was to allow markup in the description when the `itemDescriptionLength` was set to 0. It can't support markup in other situations since it has the potential of truncating in the middle of markup, like the middle of a paragraph or something, which would break the markup. So it's preferable to use the `itemContentField` field for markup. But I've updated it so that it's more clear that it allows markup when itemDescriptionLength is 0, and it's also been updated to convert relative to absolute URLs when the description contains HTML, like you suggested. Previously it was only doing this for the content field.

Btw, as far as I can tell, I don't have a completed CLA from you, so I can't use submitted code directly and don't know what your GitHub email is to attribute. If possible, please complete the short CLA form at the link below and make sure to use your GitHub email so that I can link it in PR commits: https://processwire.com/about/license/cla/ Thanks.

PR #254, opened 2023-01-15
Forpage support for showif

adrianbj:
This adds the changes proposed here: https://processwire.com/talk/topic/26943-field-dependency-on-field-within-repeater-based-on-field-outside-of-repeater/?do=findComment&comment=229807

Not sure if this covers all the requirements, but is hopefully a decent starting point for you.

Thanks for considering as this is a major pain point for me at the moment.


adrianbj:
Sorry this includes too many other changes. New PR coming soon.

PR #255, opened 2023-01-15
Add support for `forpage` in showIf

adrianbj:
Hi @ryancramerdesign - this adds the solution proposed here: https://processwire.com/talk/topic/26943-field-dependency-on-field-within-repeater-based-on-field-outside-of-repeater/?do=findComment&comment=229807

I am not sure if this is the best approach or if it handles all situations, but it works for my current needs and hopefully gives you an idea of what is needed.

Thanks,
Adrian

### DIFF
--- a/wire/templates-admin/scripts/inputfields.js
+++ b/wire/templates-admin/scripts/inputfields.js
@@ -1481,6 +1481,9 @@ function InputfieldDependencies($target) {
 			var match = part.match(/^[,\s]*([_.|a-zA-Z0-9]+)(=|!=|<=|>=|<|>|%=)([^,]+),?$/);
 			if(!match) continue;
 			var field = match[1];
+			if(field.indexOf('forpage.') == 0) {
+				field = field.replace('forpage.', '').replace(/\_repeater.*/g, '');
+			}
 			var operator = match[2];
 			var value = match[3];
 			var subfield = '';
--- a/wire/templates-admin/scripts/inputfields.js
+++ b/wire/templates-admin/scripts/inputfields.js
@@ -1482,7 +1482,7 @@ function InputfieldDependencies($target) {
 			if(!match) continue;
 			var field = match[1];
 			if(field.indexOf('forpage.') == 0) {
-				field = field.replace('forpage.', '').replace(/\_repeater.*/g, '');
+				field = field.replace('forpage.', '').replace(/\_repeater\d+/g, '');
 			}
 			var operator = match[2];
 			var value = match[3];


adrianbj:
@ryancramerdesign - the one problem I am having is that this only works if the repeater items aren't AJAX loaded. I think it's because `setupConditions()` isn't called when creating / opening a repeater item via AJAX. Any thoughts on how it could be implemented to work with AJAX loaded items as well?

Thanks.

gideonso:
I am late but I think this is very useful.

ryancramerdesign:
Thanks @adrianbj I have added this. Also, I found that it also works with ajax-loaded repeaters in testing here. So maybe something has changed since the time this PR was submitted? Or maybe my test case is too simple? My test was for an image field in a repeater and it has the show-if condition of `forpage.headline!=test`. When I set my headline field to "test" the image in the repeater item disappears, whether it was ajax loaded or not, and whether I made the change to the headline before or after the repeater item loading. Let me know if there's a case where it isn't working? I did test and find that InputfieldsInit() -- InputfieldDependencies() -- setupDependencyField() -- setupConditions() is being called after an ajax repeater item loads.

adrianbj:
Thanks @ryancramerdesign - my testing with it now shows it's working with AJAX loaded repeaters as well, so looks like the `setupConditions()` might be working differently than before?

elabx:
Hi @ryancramerdesign! Thanks for this great update! 

Would it be possible to add support for nested repeaters? I have made a proof of concept here:

https://github.com/elabx/processwire/pull/1/files

I have tested it with up to three levels of repeaters such as:

`forpage.forpage.forpage.forpage.title`

And it seems to work! Hope it  is of any help. It'd be great if anybody else can give it a test. Thanks!

Toutouwai:
It's cool to have this feature in PW. Thanks @adrianbj!

I think "forpage" was probably not the right keyword to use though, because in repeater terminology "forpage" means the page that contains the current repeater item, and this is not necessarily the page that's open in Page Edit, e.g. if a repeater is nested inside another repeater, or is inside a FieldsetPage. I think "page" would have been better (to make it consistent with the keyword used in the dependent selects feature), or maybe "editpage". Then "forpage" could be used more correctly as @elabx is suggesting.

I expect it's too late to change now as it would be breaking for existing usages?

adrianbj:
@Toutouwai - sorry, you're right - I didn't think through the nesting aspect of this and honestly it was quickly thrown together as a proof of concept fully expecting Ryan to do it differently.

Hopefully @ryancramerdesign has some thoughts about the best way forward and the ability to support nesting as @elabx suggested.

PR #256, opened 2023-01-19
When template has `noLang` set, setLanguageValue should always use the default language

MrSnoozles:
Possible fix to https://github.com/processwire/processwire-issues/issues/1118

When a multilanguage field should be in one language only for a specific template (`noLang=1`) the value should be always written into the default language.


ryancramerdesign:
@MrSnoozles Thanks for the PR. I'll close this since we fixed this over in the processwire-issues repo.

PR #259, opened 2023-02-06
use Sanitizer's email allowIDN option for email fields

poljpocket:
According to the [forum discussion](https://processwire.com/talk/topic/28125-emails-in-idn-format-do-not-get-saved-cannot-be-entered) about this topic, this PR adds a configuration field to FieldtypeEmail which allows to use Sanitizer::email's new IDN options for email addresses.

Associated issue: https://github.com/processwire/processwire-issues/issues/1680

ToDo:

- [X] add db config field and backend validation
- [X] solve problem that browsers won't allow IDN characters in local part of email address ([see W3C bug report here](https://www.w3.org/Bugs/Public/show_bug.cgi?id=15489))
- [ ] redact field configuration strings

### DIFF
--- a/wire/modules/Fieldtype/FieldtypeEmail.module
+++ b/wire/modules/Fieldtype/FieldtypeEmail.module
@@ -10,7 +10,10 @@
  * 
  * ProcessWire 3.x, Copyright 2020 by Ryan Cramer
  * https://processwire.com
+ * 
+ * @property int allowIDN Allow email in IDN format (0=no, 1=allow domain, 2=allow domain and local)
  *
+ * @see Sanitizer::email() for more info about allowIDN
  *
  */
 class FieldtypeEmail extends FieldtypeText {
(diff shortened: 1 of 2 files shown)


gebeer:
As for client side email validation with regex there is an IETF draft here: https://datatracker.ietf.org/doc/html/draft-seantek-mail-regexen-00
not sure though whether the regexes take RFC 6530 into account.

poljpocket:
Thanks @gebeer. There is a way to provide a pattern for validation even without a JS library. I will add this for testing.

poljpocket:
The pattern used is: `[^@]+@[^\.]+\...+`

It simply checks for any character(s) followed by an `@` followed by any length of a domain and a TLD with at least two characters.

ryancramerdesign:
@poljpocket Thanks, I have added something similar to this. I'd added in the IDN support and sat on it for a bit before I saw your PR, and then updated it before the commit to add some of your ideas into it as well. Please let me know if you find I missed anything important.

poljpocket:
Thanks for considering @ryancramerdesign. I have looked through your code. I think your change to not allow IDN for the system field is very good!

I am not too proud of myself about the `type="text"` change but as a temporary workaround, I think we have no choice.

PR #261, opened 2023-02-16
Added 'page-edit-redirects' optional permission

chriswthomson:
If the user role has this permission, they can add/edit redirect URLs in the page editor.


ryancramerdesign:
Thanks @chriswthomson I have added this PR

PR #262, opened 2023-02-17
Fix potentially missed else conditionals

netcarver:
Following a previously reported missed 'else' conditional, here are two more potential cases.

### DIFF
--- a/wire/core/PagesPathFinder.php
+++ b/wire/core/PagesPathFinder.php
@@ -1424,7 +1424,7 @@ class PagesPathFinder extends Wire {
 			$template = $this->getResultTemplate();
 			if(!$template) {
 				return false; // may need to detect later
-			} if(in_array($template->name, $config->adminTemplates, true)) {
+			} else if(in_array($template->name, $config->adminTemplates, true)) {
 				$this->admin = true;
 			} else if(in_array($template->name, array('user', 'role', 'permission', 'language'))) {
 				$this->admin = true;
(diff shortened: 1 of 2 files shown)


wire/core/PagesPathFinder.php
```diff
@@ -1424,7 +1424,7 @@ protected function isResultInAdmin() {
 			$template = $this->getResultTemplate();
 			if(!$template) {
 				return false; // may need to detect later
-			} if(in_array($template->name, $config->adminTemplates, true)) {
+			} else if(in_array($template->name, $config->adminTemplates, true)) {
```
netcarver, line 1427: This one probably has no effect.

netcarver:
@ryancramerdesign Ryan, there's another instance of this in ProCache as well. A quick grep shows this - but I've not checked the context so it might be a false positive.

```
public_html/site/modules/ProCache/ProCacheStaticEntry.php:170:		} if($key === 'created') {
```

netcarver:
Here's a `missingelse.yaml` semgrep rule for finding these in .php files which you can add to your QA pipeline

```yaml
rules:
- id: missing-else-conditional
  languages:
    - php
  severity: WARNING
  pattern-regex: "}[ \\t]*if\\s*\\("
  message: Found possible missing else conditional
```

netcarver:
Agh, ignore that last commit, Ryan, made by mistake.

ryancramerdesign:
Thanks @netcarver I have added this

netcarver:
@ryancramerdesign Thank you, Ryan. Will close this now. 

BTW, was that also a missed else in ProCache?

ryancramerdesign:
Yep, the one is procache didn't matter in terms of the logic but fixed it
still to be clean. Thanks

PR #266, opened 2023-03-22
add possibility to open PageTable items in new tab

jmartsch:
This simple change enables us to open the link of a PageTable item in a new tab via right click or middle click instead of a modal dialog. The normal modal dialog behaviour is still working if you use a simple click.


ryancramerdesign:
Thanks @jmartsch I have added this in as part of some other FieldtypePageTable updates

PR #268, opened 2023-04-06
fix: Response Code 400 with UTF-8 in URI

gebeer:
see https://github.com/processwire/processwire-issues/issues/1718


matjazpotocnik:
@gebeer PW issue #1718 is now closed, so perhaps you could close this one too?

gebeer:
> @gebeer PW issue #1718 is now closed, so perhaps you could close this one too?

Thanks for the reminder

PR #270, opened 2023-04-29
Update WireInput.php

netcarver:
Add missing ' to correct inversion of syntax highlighting in the documentation page for cookie().

Ref: https://processwire.com/api/ref/wire-input/cookie/


ryancramerdesign:
Thanks @netcarver I have added this.

PR #271, opened 2023-05-04
Use the same mock/placeholder Page object for custom fields at runtime

romaincazier:
This is a bit specific but this PR allows to use the same mock/placeholder Page object (per file) for custom fields at runtime.

This bit came up as an issue when trying to use the 3rd party module [FieldtypeOembed](https://github.com/neuerituale/FieldtypeOembed) (which extends FieldtypeURL) as a custom field. To avoid repetitive calls to get remote oembed data the result is cached in the db and later set as a property of the field’s page on [line 115](https://github.com/neuerituale/FieldtypeOembed/blob/64cf553cc45eeb51d72afaad09452057b298e053/FieldtypeOembed.module#L115).

However between [`___wakeupValue()`](https://github.com/neuerituale/FieldtypeOembed/blob/64cf553cc45eeb51d72afaad09452057b298e053/FieldtypeOembed.module#L110) and [`___formatValue()`](https://github.com/neuerituale/FieldtypeOembed/blob/64cf553cc45eeb51d72afaad09452057b298e053/FieldtypeOembed.module#L149) a new dummy `$page` has been created and thus is not holding the cached value anymore.

This PR solves this by using the same placeholder Page per file throughout the rendering process.

I don’t think there are any downsides to this, though I could be wrong.

Thanks for considering.

### DIFF
--- a/wire/core/Pagefile.php
+++ b/wire/core/Pagefile.php
@@ -95,6 +95,14 @@ class Pagefile extends WireData implements WireArrayItem {
 	protected $fieldValues = array();
 
 	/**
+	 * Reference to the mock/placeholder Page object used for custom fields
+	 * 
+	 * @var Page
+	 * 
+	 */
+	protected $filePage;
+
+	/**
 	 * Created user (populated only on rquest)
 	 *
 	 * @var User|null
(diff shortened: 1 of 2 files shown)


romaincazier:
See https://github.com/neuerituale/FieldtypeOembed/issues/5 for more context

romaincazier:
Hi @ryancramerdesign,

I just tested your suggestion and in this specific case this won’t cut it as the issue is happening earlier in the process (difference between `wakeupValue` and `formatValue`). However I do agree this is something that could be solved on the module’s side, which I actually did and [made a pull request for](https://github.com/neuerituale/FieldtypeOembed/pull/7).

If you think this is not worth it, we can close this PR.

Thanks for considering.

PR #273, opened 2023-06-12
Make error message strings translatable in ProcessPageAdd

jmartsch:


ryancramerdesign:
Thanks @jmartsch I have added this

PR #274, opened 2023-06-20
Bug Fix: InputfieldPageAutocomplete.js can't trigger a requiredIf

hiboudev:
In InputfieldPageAutocomplete.js, value always start with a comma, so inputfields.js->inputfieldChange() doesn't parse it correctly. So other fields that depends on it by a requiredIf are never updated.


hiboudev:
@ryancramerdesign I'm using this fix since months and didn't see any issue so far.

PR #275, opened 2023-07-19
Access check on multiple templates

chriswthomson:
Prevents templates appearing in the 'Add New' menu that the user role does not have permission to add.


ryancramerdesign:
@chriswthomson I think this could potentially be a lot of overhead on a larger site, as this particular loop would find all possible parent pages for all templates. It may not matter on a smaller site, but on some larger sites I'm not sure it could scale. It's a good idea though, so I'll have to keep looking at it. Thanks.

PR #276, opened 2023-08-09
Pageimage: Added webpOptions override

mpsn:
Added Parameter:
Pageimage::webp(optional $webpOptions)

This allows adjusting webp quality on per-variant basis like the following:
```php
$webpUrl = $img->webp(["quality"=>70])->url;
```
instead of
```php
global $config;
$webpConf = $config->get('webpOptions');
$webpConf["quality"] = 70;
$config->set('webpOptions', $webpConf);
$webpUrl = $img->webp()->url;
```
or manually creating different image objects.

### DIFF
--- a/wire/core/Pageimage.php
+++ b/wire/core/Pageimage.php
@@ -1702,17 +1702,19 @@ class Pageimage extends Pagefile {
 	/**
 	 * Get WebP "extra" version of this Pageimage
 	 *
+	 * @param array $webpOptions Optionally override image webp options.
 	 * @return PagefileExtra
 	 * @since 3.0.132
 	 *
 	 */
-	public function webp() {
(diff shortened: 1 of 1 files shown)


teppokoivula:
> Yeah, should be good enough.
> 
> I guess that's a trait shared among all image quality settings, unless image file names are reworked. 🙃

A bit off-topic, but for this reason https://processwire.com/modules/unique-image-variations/ is, in my opinion, a must-have module. Doesn't do anything to WebP variations, though. Perhaps that could be added.

mpsn:
If webp were treated as fully-fledged image variants instead of extras the different naming scheme would apply as well.
It might be worth eliminating the _scale original format_->_generate webp extra_-workflow anyway, even more if PW core eventually adds additional file support (like AVIF).

PR #277, opened 2023-09-14
No more eval in templates-admininputfields.js

mpsn:
Non-lax javascript security settings or proper CSP tend to block string eval, like found in templates-admin/scripts/inputfield.js.

(The same issue exists within FormBuilder/form-builder.js)


ryancramerdesign:
Thanks @mpsn I have added this and will update the FormBuilder file for the next version as well.

PR #278, opened 2023-11-10
Add hookable ImageSizerEngineIMagick::imSaveReady() method

Toutouwai:
For equivalence with ImageSizerEngineGD, so that third-party modules may access the uncompressed image data before a resize regardless of which core ImageSizerEngine is selected. This will allow for the saving of images in high-efficiency compressed formats such as AVIF, and other formats when they are introduced in the future.


ryancramerdesign:
@Toutouwai Thanks, I have added this

PR #279, opened 2023-11-27
Update ParsedownExtra.php

erikmh:
To work around deprecated 'Use of "self" in callables' in PHP 8.2. See `https://www.designcise.com/web/tutorial/how-to-fix-use-of-self-in-callables-is-deprecated-php-error`


ryancramerdesign:
Thanks @erikmh I have added this.

PR #280, opened 2024-02-06
Update InputfieldSelector.js to work with repeaters

elabx:
Inputfield selector doesn't work if loaded through a repeater.


elabx:
You are super right, just saw the bugs in a bunch of my sites just today. Maybe you have a reference at hand from another Inputfield on how to implement this? If not, that's ok too, I've got enough clues on your reply! Appreciate your time and attention on this matter!

PR #282, opened 2024-03-15
Fix PHP and SQL timezone mismatch when saving new page: set created, modified, published dates with PHP instead of NOW()

donatasben:
To prevent difference between PHP and MySQL server timezones when creating new pages I propose to always use PHP to set created/modified/published timestamps.

Reason: it can become very tricky to compare new pages' `created` dates using php `date()` and similar as they might become offset by several hours if SQL server timezone is different than PHP's (or that which PW config is set to) as the `created` is then set with SQL timezone. it would be required to know SQL timezone to correctly correct dates for comparison.

Hope there was no very important reason why decision was made to use SQL's "NOW()" when creating new pages.

### DIFF
--- a/wire/core/PagesEditor.php
+++ b/wire/core/PagesEditor.php
@@ -527,6 +527,7 @@ class PagesEditor extends Wire {
 		$userID = $user ? $user->id : $config->superUserPageID;
 		$systemVersion = $config->systemVersion;
 		$sql = '';
+		$sqlNowCustom = '"'.date('Y-m-d H:i:s').'"'; // Use instead of MySQL's NOW() to avoid timezone difference with PHP
 		
 		if(!$page->created_users_id) $page->created_users_id = $userID;
 		
@@ -570,7 +571,7 @@ class PagesEditor extends Wire {
 		}
 
 		if(empty($options['quiet'])) {
-			$sql = 'modified=NOW()';
+			$sql = 'modified='.$sqlNowCustom;
 			$data['modified_users_id'] = (int) $userID;
 		} else {
 			// quiet option, use existing values already populated to page, when present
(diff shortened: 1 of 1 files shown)


adrianbj:
Not sure what other folks do, but whenever I set up a new server I make sure that system, MySQL and PHP timezones match. This means that things will always match no matter how the time is set.

BernhardBaumrock:
@adrianbj did you suggest @ryancramerdesign that ProcessWire does a check for that and shows a warning somewhere if there is a mismatch?

BernhardBaumrock:
@donatasben yes of course I'm talking about a runtime check, not during installation as this would be different for local DDEV installations and when deploying the site afterwards to staging/production.

I've asked my AI and this is what it came up with:

```php
// Get the current timezone offset in hours from MySQL
$mysqlTimezoneOffset = wire('database')->query("SELECT TIMESTAMPDIFF(HOUR, UTC_TIMESTAMP, NOW())")->fetchColumn();

// Get the current timezone offset in hours from PHP
$phpTimezoneOffset = date('Z') / 3600; // 'Z' returns timezone offset in seconds

if ($mysqlTimezoneOffset == $phpTimezoneOffset) {
    echo "MySQL and PHP timezones have the same offset: $phpTimezoneOffset hours";
} else {
    echo "MySQL timezone offset: $mysqlTimezoneOffset, PHP timezone offset: $phpTimezoneOffset";
}
```

For my server that worked. So wouldn't it be good if the PW backend showed a warning if there is a mismatch?

BernhardBaumrock:
PS: I understand that this does not solve your issue if you are not able to set the timezone for your server/mysql but I think it would be good to have nevertheless :)

PR #287, opened 2024-05-09
Allow setting vendor/ location from an ENV variable

netcarver:
Defaults to existing vendor/ location, but allows this to be overridden by reading the PW_VENDOR_PATH environment variable.

For example, setting PW_VENDOR_PATH to ../vendor/ allows composer to install code outside the site root.

See
- https://processwire.dev/integrate-composer-with-processwire/


BernhardBaumrock:
@netcarver is on fire :) Thx for all your contributions!!

PR #288, opened 2024-05-15
Bug fix: handle invalid SVG markup

daun:
Fixes processwire/processwire-issues#1925

Only try to read the width and height attributes from an svg image's root if the parser actually returns a result. Not doing this results in a fatal error if the svg markup is invalid and the simplexml handler returns `false`.


daun:
Is anything required to get this merged? Let me know if it needs more documentation.

The issue in question is breaking a production site and I'd love to see this released.

ryancramerdesign:
Thanks @daun this has been added

PR #299, opened 2024-07-13
Allow custom markup to be set per instance

Toutouwai:
Related forum post: https://processwire.com/talk/topic/30194-weekly-update-%E2%80%93%C2%A011-july-2024/?do=findComment&comment=242758

> Another admin customisation issue I've struck recently is when you want to customise markup for InputfieldWrapper. There's [InputfieldWrapper::setMarkup()](https://github.com/processwire/processwire/blob/e508cfa2a74971490d6045bd5eb6f2a198d37037/wire/core/InputfieldWrapper.php#L1825-L1835) but this is marked as internal and isn't that practical to use because InputfieldWrapper::markup is static. That means any custom markup you set there persists for all InputfieldWrappers that are subsequently rendered, when what I want to do is customise markup for a specific instance of InputfieldWrapper.

### DIFF
--- a/wire/core/InputfieldWrapper.php
+++ b/wire/core/InputfieldWrapper.php
@@ -103,6 +103,8 @@ class InputfieldWrapper extends Inputfield implements \Countable, \IteratorAggre
 	);
 
 	static protected $markup = array();
+	
+	protected $instanceMarkup = array();
 
 	/**
 	 * Classes used during the render() method - customize with InputfieldWrapper::setClasses($array)
@@ -715,7 +717,7 @@ class InputfieldWrapper extends Inputfield implements \Countable, \IteratorAggre
 		$columnWidthSpacing = $this->getSetting('columnWidthSpacing');
 		$quietMode = $this->getSetting('quietMode');
 		$lastInputfield = null;
-		$_markup = array_merge(self::$defaultMarkup, self::$markup);
+		$_markup = array_merge(self::$defaultMarkup, self::$markup, $this->instanceMarkup);
 		$_classes = array_merge(self::$defaultClasses, self::$classes);
 		$markup = array();
 		$classes = array();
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
@Toutouwai Wouldn't this only affect the markup from that one instance of InputfieldWrapper? Forms usually have multiple InputfieldWrappers in them, which is why this is stored statically and set with `InputfieldWrapper::setMarkup()`. 

Are there cases where you'd be rendering multiple forms using different wrapper markup in the same request? I've not come across that yet. But I suppose you could do it this way: 
~~~~~
// save current markup
$originalMarkup = InputfieldWrapper::getMarkup();

// set new custom markup
InputfieldWrapper::setMarkup([ ... ]); 

// render form
echo $form->render();

// restore original markup
InputfieldWrapper::setMarkup($originalMarkup);
~~~~~
That whole thing could potentially be abstracted behind the `$form->render()`, maybe with a `$form->renderCustom($markup)` or something like that?

Toutouwai:
> Wouldn't this only affect the markup from that one instance of InputfieldWrapper?

Yes, that's exactly the aim of the pull request.

For me this request is about the ability to customise specific parts of the admin markup via hooks without affecting other InputfieldWrappers.

As an example, I wanted to add some markup adjacent to (not within) the `<label>` element of items within a repeater field.

With the changes in this pull request I can do this:

```php
$wire->addHookBefore('InputfieldWrapper(name^=repeater_form_)::render', function(HookEvent $event) {
	$wrapper = $event->object;
	$markup = [
		'item_label' => "<label class='InputfieldHeader ui-widget-header{class}' for='{for}'>{out}</label><div class='my-extra-mark'>My extra markup here</div>",
	];
	$wrapper->setInstanceMarkup($markup);
});
```



To achieve these targeted changes it's not sufficient to set the static $markup property before the form render and reset it after the form render because all the InputfieldWrappers within each repeater item will be affected before $markup can be reset.

PR #300, opened 2024-08-30
restructure install.php to do away with warnings

poljpocket:
Summary:
Restructure install.php to do away with "headers already sent" warnings because PW's session gets initialized too late

Details:
The warnings in question happen on modern PHP versions in the last step of the installer. In this step, ProcessWire's index.php file is included. But this currently happens after some of the installer's HTML already has been output. This PR changes the inclusion of the file to happen before any output happens.

This issue has been discussed in the forums: https://processwire.com/talk/topic/30321-consistent-errors-during-installation-does-anyone-else-get-these

### DIFF
--- a/install.php
+++ b/install.php
@@ -98,29 +98,33 @@ class Installer {
 		// these two vars used by install-head.inc
 		$title = "ProcessWire " . PROCESSWIRE_INSTALL . " Installer";
 		$formAction = "./install.php";
-		
-		require("./wire/modules/AdminTheme/AdminThemeUikit/install-head.inc"); 
 
-		$step = $this->post('step');
+        $step = $this->post('step');
+
+        if ($step !== null) {
+            $step = (int) $step;
+        } else {
+            $step = -1;
+        }
+
+        /** @var ProcessWire $wire */
+
+        if ($step === 5) {
+            require("./index.php");
+            $wire->modules->refresh();
+        }
 		
-		if($step === null) {
-			$this->welcome();
-		} else {
-			$step = (int) $step;
-			switch($step) {
-				case 0: $this->initProfile(); break;
-				case 1: $this->compatibilityCheck(); break;
-				case 2: $this->dbConfig();  break;
-				case 4: $this->dbSaveConfig();  break;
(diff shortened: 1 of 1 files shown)


poljpocket:
@ryancramerdesign sorry to bump this PR this way but wouldn't that be a great addition for the new upcoming release?

ryancramerdesign:
Thanks @poljpocket but I have't been able to duplicate the issue. I run the installer all the time, but have never seen this particular issue, so am unsure how to observe it?

poljpocket:
@ryancramerdesign for me, it happens every time I install ProcessWire on PHP8.0+ specifically **enable debug mode** in I think step 4 of the installer. If you install PW and leave the debug mode switch on default (off), there are no warnings.

poljpocket:
> @ryancramerdesign for me, it happens every time I install ProcessWire on PHP8.0+ when I specifically **enable debug mode** in I think step 4 of the installer. If you install PW and leave the debug mode switch on default (off), there are no warnings.

Which is something I always do because most of my fresh PW installs will be local devlopment environments inside Docker.

ryancramerdesign:
@poljpocket Strange, PHP 8.2 here and always start with debug mode on as well, and have never seen the error. Just tested again now to be sure. Maybe something to do with different PHP settings or something? Anyway, looking at the suggested change, it makes sense that the install-head.inc file should be included after the index.php file, so I will add this change, or something close to it. Thanks.

poljpocket:
You can easily reproduce my issues with it by running the installer with this: https://github.com/poljpocket/processwire-docker. The README in the repo has the instructions on how to do it.

My custom PW Docker image uses PHP 8.2 with a very simple php.ini which you can find here https://github.com/poljpocket/wordpress-docker-image/blob/main/config/php.ini. One of the consequences of that might be that it uses a particularly verbose error and warning setting. Which is good for development and actually the only purpose of the Docker image.

However, others have similar issues on non-Docker environments (see the forum post).

But as you already noted, it makes sense to include `index.php` _before_ `install_head.inc` no matter the error verbosity :).

Thanks for the effort!

PR #303, opened 2024-09-14
Optimization: also cache the absence of value; 

tuomassalo:
fixes https://github.com/processwire/processwire-issues/issues/1970


tuomassalo:
Already fixed. https://github.com/processwire/processwire-issues/issues/1970#issuecomment-2466300525

PR #306, opened 2024-10-22
show class on modulespage

BernhardBaumrock:
## Before PR:

image

## The problem

Find the module "FieldsetPage" or the module "FieldtypeFloat".

This is especially (but not only) a problem when having a non-default language active. How should I ever know that the translation lists "FieldtypeFloat" under "Fließkomma-Zahl" and not "Fließkommazahl". Or how should I know that "Fieldset (Close)" is not translated yet, but "Fieldset (Page)" is translated to "Fieldset (Seite)".

The problem with the tooltips is that you can't simply use CMD+F for finding their names via browser!

I wasted so much time today finding "FieldsetPage" that it justified creating a PR 😄 

## After PR:

image

Now you can search for "FieldtypeFloat" and your browser will show you where the module is!

### DIFF
--- a/wire/modules/Process/ProcessModule/ProcessModule.module
+++ b/wire/modules/Process/ProcessModule/ProcessModule.module
@@ -881,7 +881,7 @@ class ProcessModule extends Process {
 			$info = $modules->getModuleInfoVerbose($name); 
 			$configurable = $info['configurable'];
 			$title = !empty($info['title']) ? $sanitizer->entities1($info['title']) : substr($name, strlen($section));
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
@BernhardBaumrock There's already an option for that, it's the [allowClasses](https://github.com/processwire/processwire/blob/master/wire/modules/Process/ProcessModule/ProcessModule.module#L793) option to that method. By default it's only used on the new modules screen, but what you suggest makes sense to me, I'll make it the default setting so that it shows classes with all the modules.

BernhardBaumrock:
Great, thank you :)

PR #307, opened 2024-11-08
Add helpful information for why module cannot be uninstalled

artfulrobot:
When trying to uninstall a module you get a message if you can't because other modules depend on it, or because a field uses it. But you're not told which one. As this info is not listed anywhere else, it's quite hard to find out why.

This PR simply adds to the explanation text:

- the list of modules for which the module-to-be-uninstalled is a dependency.
- the list of the (first) field that uses the module.


wire/core/ModulesInstaller.php
```diff
@@ -266,14 +266,14 @@ public function isUninstallable($class, $returnReason = false) {
 				$reason = $this->_("Module is permanent");
 			} else {
 				$dependents = $this->getRequiresForUninstall($class);
-				if(count($dependents)) $reason = $this->_("Module is required by other modules that must be removed first");
+				if(count($dependents)) $reason = $this->_("Module is required by other modules that must be removed first: "  . implode(', ', $dependents));
```
teppokoivula, line 269: Note: this is a translatable string. For use cases like this you’ll want to use sprintf() or add the list of values after the language method call.

artfulrobot: Ah, good point, Thanks I will adjust PR.

artfulrobot: Resolved, I think by https://github.com/processwire/processwire/pull/307/commits/ffdf9f2a438238a7d9d2af37042a14f6b5b43faf

wire/core/ModulesInstaller.php
```diff
@@ -266,14 +266,20 @@ public function isUninstallable($class, $returnReason = false) {
 				$reason = $this->_("Module is permanent");
 			} else {
 				$dependents = $this->getRequiresForUninstall($class);
-				if(count($dependents)) $reason = $this->_("Module is required by other modules that must be removed first");
+				if(count($dependents)) $reason = $this->_(sprintf(
+					"Module is required by other modules that must be removed first: %s",
+					implode(', ', $dependents))
+				);
```
poljpocket, line 272: This will not work correctly for translations. Just the string including the format should be translatable:

```php
if(count($dependents)) $reason = sprintf(
	$this->_("Module is required by other modules that must be removed first: %s"),
	implode(', ', $dependents)
);
```

wire/core/ModulesInstaller.php
```diff
@@ -266,14 +266,20 @@ public function isUninstallable($class, $returnReason = false) {
...
 
 			if(!$reason && in_array('Fieldtype', wireClassParents($namespace . $class))) {
 				foreach($this->wire()->fields as $field) {
 					$fieldtype = wireClassName($field->type, false);
 					if($fieldtype == $class) {
-						$reason = $this->_("This module is a Fieldtype currently in use by one or more fields");
+						$reason = $this->_(sprintf(
+							"This module is a Fieldtype currently in use by one or more fields including: %s",
+							$field->name)
+						);
```
poljpocket, line 282: Same here, see above:

```php
$reason = sprintf(
	$this->_("This module is a Fieldtype currently in use by one or more fields including: %s"),
    $field->name
);
```

poljpocket (changes requested):
Your sprintf changes don't really work correctly right now.

wire/core/ModulesInstaller.php
artfulrobot: Thanks for the engagement, I'll update the pr soon.

PR #316, opened 2025-02-10
add comments to avoid confusion how to add hooks in admin.php

BernhardBaumrock:
This has bit me myself and I'm not the only one: https://github.com/processwire/processwire-issues/issues/2037#issuecomment-2647757222

### DIFF
--- a/site-blank/templates/admin.php
+++ b/site-blank/templates/admin.php
@@ -14,5 +14,8 @@
  * 
  */
 
+// PLACE YOUR HOOKS HERE
+
 /** @var Config $config */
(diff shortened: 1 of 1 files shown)


poljpocket:
Isn't this specified in the file already?

https://github.com/processwire/processwire/blob/44fcf13ea2d7f14a04eed54c29afcc79eb46ec45/site-blank/templates/admin.php#L11

`at the end` specifically.

BernhardBaumrock:
Yes, but it could be made more obvious and thus help prevent issues and that's what I suggest here.

PR #318, opened 2025-04-01
FIX: Check $test is not null before use in strpos() for Notices 

brodiefarrell:
In `Notices.php`, there is a chance that `$text` (derived from `$item->text`) can be used in strpos whilst being null. This causes a warning in PHP 8.


hiSandog:
Would it be safer to normalize or reject `null` when a `Notice` is constructed, rather than guarding only this `strpos()` call? A focused test could create a notice with `null` text and exercise string conversion, duplicate detection, and rendering to establish the intended contract. That would address the other null-sensitive paths already identified without scattering inconsistent checks.

PR #322, opened 2025-07-10
Proposal: A Core Hook System for a More Modular and Extensible JavaScript

BernhardBaumrock:
Hi Ryan,

I'm excited to propose a new feature for the core: a JavaScript hook system that mirrors the power and elegance of our beloved PHP hooks.

This module introduces addHookBefore(), addHookAfter(), and event.replace to any JavaScript object or class. It's designed to make our client-side code as modular, extensible, and fun to work with as the PHP side. Imagine third-party modules being able to cleanly and reliably modify the behavior of core UI components, like modals or page tree actions.

I've included a full set of interactive examples (including an Alpine.js demo) in the ProcessJavaScriptHooks module to showcase how it works.

This could be a huge step forward for ProcessWire's client-side architecture. I'd be thrilled for you to take a look and consider it for the core.

Thanks!
Bernhard

PS: You can either checkout this PR directly or install the standalone module that I have created for development here: https://github.com/baumrock/JavaScriptHooks

### DIFF
--- a/wire/modules/JavaScriptHooks/examples/06-texttags.php
+++ b/wire/modules/JavaScriptHooks/examples/06-texttags.php
@@ -19,11 +19,16 @@
         }
       }
 
-      // add TextTags() as global function
-      window.TextTags = function(text) {
+      const func = function(text) {
         const instance = ProcessWire.wire(new TextTags());
         return instance.log(text);
-      };
+      }
+
+      // add TextTags() as global function
+      window.TextTags = func;
+
+      // add TextTags() as ProcessWire.TextTags() function
+      ProcessWire.TextTags = func;
     })();
 
     // add demo hook to modify the log() behavior
@@ -33,7 +38,9 @@
     });
 
     TextTags("Hello, world!");
+
+    ProcessWire.TextTags("Hi there from ProcessWire.TextTags()");
   </script>
 
-  <p>Example built as response to <a href=https://github.com/baumrock/processwire/pull/1>https://github.com/baumrock/processwire/pull/1</a></p>.
+  <p>Example built as response to <a href=https://github.com/baumrock/processwire/pull/1>https://github.com/baumrock/processwire/pull/1</a>.</p>
 </div>


BernhardBaumrock:
Hey @poljpocket absolutely valid points. I didn't mention this and it's one of the main reasons for being in the core:

If it is not, the core will never use it. And I think it should.

If it is in the core, then the core will use it and so can any other JS based module. It's been a tremendously helpful pattern for me in RockCommerce, RockGrid, etc.; You'll only understand once you start adopting to it. And it will never happen if it's a module that you have to install first.

I did implement it as a module (RockJavaScriptHooks) first to make sure it works. And it works extremely well!

I truly believe this will have a very positive impact on ProcessWire. Think of it that way: What if ProcessWire didn't have the PHP hook system. Would it work? For sure! Would it be possible to customise things? For sure - one way or another.

The key takeaway is the following for anybody developing anything based on JS for ProcessWire: Want to make something customisable by the user? Just add three underscores and that's it!

Does that answer your questions?

BernhardBaumrock:
Oh, and I somewhat disagree with your statement that it only adds value if the core uses it! Think about url hooks. They have not been there until they have been added. And when they have been added nothing in the core was using url hooks. Have they been a valuable addition? In my opinion yes!

We could have kept hooking into the 404 process, but url hooks are so much nicer!

poljpocket:
Ok, now I think you misunderstood.

What I'm saying is: Shouldn't we modify the JavaScript code in the core to use it and add real-world examples directly in this PR? What does this new addition help if the core JavaScript isn't hookable?

I think your examples don't get the point across. It's cool to have a couple of hello-world examples. But that increment in Alpine.js is only useful if the buttons and the first part of the code is actually part of the core and thus something you won't be able to modify without changing the core JavaScript files.

But here are examples which come to mind:

- Extend the select field to allow for AJAX-driven dropdown options.
- Extend the text-tags field to allow for color-coded items.

I can help if you want. Because I think this is very cool and useful too.

poljpocket:
Hi everyone. Yes, let's continue the discussion here outside of my PR of Bernhard's PR 😄 

@ryancramerdesign I think it's important to point out that this proposal has been misunderstood multiple times by me and others: Just to reiterate, to include this in the core, all JavaScript will need to be changed or rewritten because right now, the code isn't designed at all with hooks in mind. Bernhard's proposal isn't at all how the code would look like because it doesn't actually change any code in the core. If he wants to for example make his RookCommerce module hookable, he can do that right now with this implementation. He can do that without changing a line of code in the core. As long as he isn't relying on any fields from the core that is.

@BernhardBaumrock That's exactly why I said "your examples don't get the point across" time and time again. We should go ahead and create a real-world example. Your proposal with the date picker is perfect: It's something in the core which right now, can't easily be changed using the current codebase and modules won't be able to do this at all unless the module provides a complete replacement of the datepicker component. I'll get rid of my PR-PR and let's focus on that example and get it done. So, this PR should:

1. change the JS in the core which handles the datepicker component
2. add a complete module which will alter the date selectability based on the new hook system

poljpocket:
> I want to clarify as well: All JavaScript in the core CAN stay untouched. This PR will not have any side effects to any existing JS implementations in the core. It will only kick in once a class is "wired" via ProcessWire.wire(...)

It can't because if you look at it, it's not written in any way at all which makes hooks work.

Let me make an example:

https://github.com/processwire/processwire/blob/44fcf13ea2d7f14a04eed54c29afcc79eb46ec45/wire/modules/Inputfield/InputfieldDatetime/InputfieldDatetime.js#L104-L125

It is part of the JS code which handles `InputfieldDatetime`. It validates date selections for the multiple `<select>` date input type. This code needs to be changed around completely to allow any hook to reject a certain date selection. You can't just throw your cool new `ProcessWire.wire()` around it and be done.

And as I said before, any code NOT in the core doesn't need hookability as a core concept because you the developer can just add hookability to your code by adding something similar to the `Hooks` class (also `HookEvent` and `HooksHelper`) to the code base.

BernhardBaumrock:
@poljpocket my point is that merging this PR does not mean that we have to refactor hundreds of core JS implementations. We CAN if we want to and if it adds a benefit, but we don't have to.

Of course, if we don't change any core implementations than they will not be hookable, that's obvious.

> to include this in the core, all JavaScript will need to be changed or rewritten because right now, the code isn't designed at all with hooks in mind.

I was just afraid that someone could misinterpret this as that merging my PR means that "all JavaScript will need to be changed/rewritten", which is not the case.

I'm proposing a standard that everybody can use so that we all don't have to reinvent the wheel over and over again by adding custom events that somebody can listen to. Or by adding custom translations somewhere that someone can use to modify the behaviour of a module. Or by offering callbacks somewhere that might not even be documented.

All we need is three underscores and things will be obvious, coherent, easier to maintain and more powerful. With no overhead and no obligations.

PR #327, opened 2026-04-01
Cleanup unused container classes

adrianbj:
Hey @ryancramerdesign - I think I mentioned not long ago that I thought the admin.css file seemed unnecessarily huge so I asked Claude to have go at removing redundant classes and it literally halved its size. I know I saw a note somewhere in your code about not minifying this file (although I'm not sure why), but at least this should be a big improvement.

I haven't tested it thoroughly, but a quick look around the admin looks good.

I asked Claude:

❯ How is it possible that you could simplify all those selectors so much without breaking anything or making them not specific enough so that they don't interfere with other page elements?

⏺ Great question. Let me show you exactly what happened with a concrete example. Take line 32 from the original file:

  Before:
  h1,.uk-h1,h2,.uk-h2,h3,.uk-h3,...,
  .uk-section-primary:not(.uk-preserve-color) h1,
  .uk-section-primary:not(.uk-preserve-color) .uk-h1,
  .uk-section-secondary:not(.uk-preserve-color) h1,
  .uk-tile-primary:not(.uk-preserve-color) h1,
  .uk-card-primary.uk-card-body h1,
  .uk-card-secondary.uk-card-body h1,
  .uk-overlay-primary h1,
  ... {
      color: var(--pw-text-color)
  }

  After:
  h1,.uk-h1,h2,.uk-h2,h3,.uk-h3,... {
      color: var(--pw-text-color)
  }

  The removed selectors were more specific versions of the base selectors, targeting elements inside containers like .uk-section-primary. For example, .uk-section-primary:not(.uk-preserve-color) h1 means "an h1
  inside a .uk-section-primary element".

  The key points:

  1. Those container classes don't exist in the PW admin markup — no PHP template ever outputs class="uk-section-primary" or class="uk-tile-secondary" etc. So those selectors never matched anything.
  2. The base selectors are preserved — h1, .uk-h1, a:hover, .uk-button, etc. all remain untouched. These are the ones that actually style the admin.
  3. The properties are identical — the removed selectors applied the exact same CSS properties as the base selectors. They existed as UIkit's way of ensuring correct colors inside colored containers — a UIkit feature the PW admin doesn't use.
  4. No specificity change for real elements — since the removed selectors targeted elements inside non-existent containers, they were never winning any specificity battles. The base selectors were always the ones doing the actual work.

  So nothing got "simplified" in terms of what CSS actually applies — we just removed selectors that could never match any element in the admin.

### DIFF
--- a/wire/modules/AdminTheme/AdminThemeUikit/themes/default/admin.css
+++ b/wire/modules/AdminTheme/AdminThemeUikit/themes/default/admin.css
@@ -8,724 +8,716 @@ a, .uk-link { /* RJC */
 	color: var(--pw-main-color);
 }
 
-.uk-section-primary:not(.uk-preserve-color),.uk-section-secondary:not(.uk-preserve-color),.uk-tile-primary:not(.uk-preserve-color),.uk-tile-secondary:not(.uk-preserve-color),.uk-card-primary.uk-card-body,.uk-card-primary>:not([class*='uk-card-media']),.uk-card-secondary.uk-card-body,.uk-card-secondary>:not([class*='uk-card-media']),.uk-overlay-primary {
-	color: light-dark(var(--pw-blocks-background), var(--pw-text-color));
-}
-
-.uk-section-primary:not(.uk-preserve-color) a,.uk-section-primary:not(.uk-preserve-color) .uk-link,.uk-section-secondary:not(.uk-preserve-color) a,.uk-section-secondary:not(.uk-preserve-color) .uk-link,.uk-tile-primary:not(.uk-preserve-color) a,.uk-tile-primary:not(.uk-preserve-color) .uk-link,.uk-tile-secondary:not(.uk-preserve-color) a,.uk-tile-secondary:not(.uk-preserve-color) .uk-link,.uk-card-primary.uk-card-body a,.uk-card-primary.uk-card-body .uk-link,.uk-card-primary>:not([class*='uk-card-media']) a,.uk-card-primary>:not([class*='uk-card-media']) .uk-link,.uk-card-secondary.uk-card-body a,.uk-card-secondary.uk-card-body .uk-link,.uk-card-secondary>:not([class*='uk-card-media']) a,.uk-card-secondary>:not([class*='uk-card-media']) .uk-link,.uk-overlay-primary a,.uk-overlay-primary .uk-link {
-	color: inherit;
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
I think that a lot of those extra classes may be to fully support all uikit
features, as I know Diogo put a lot of work into that. It was because
Bernhard had concerns about it initially missing some uikit
features/styling that he was using in his modules, but the core wasn't
using. So the konkat guys went and added support for the whole uikit
library, as I understand it.

I had a similar thought this week that Claude would probably be great to
work in some of processwire's css and improve it. I'm sure there's room for
improvement in just about every css, scss and less file. For instance, it
might be a good opportunity to remove the dependency between the original
uikit theme and the new default one. It seems like there's good potential
for the admin.css to be self contained and not need to override original
theme styles. I also think that Diogo has started using Claude with the
admin.css now too. But they are still sending over tweaks and improvements
(some this week even) so probably should keep things they way they are now.
Longer term I think Claude is going to be very helpful in optimizing css
throughout the core.

adrianbj:
Great point about supporting Uikit classes for third party modules. I wonder if there could be somehow be chunked out and only loaded if needed.

I was curious what Claude thought was the bigger gain:

❯ Does removing all these extra selectors speed rendering up or just loading of the CSS file.                                                                                                                      
                                                                                                                                                                                                                 
⏺ Both, but to different degrees:                                                                                                                                                                                  
                                                                                                                                                                                                                 
  Loading — the bigger win. 122KB less to download, parse, and tokenize. CSS parsing is synchronous and render-blocking, so this directly reduces time-to-first-paint.                                             
   
  Rendering — a real but smaller win. The browser builds a data structure (the style rule map) from all selectors and checks them during style recalculation. With ~2,000 fewer selectors, the browser has a       
  smaller map to maintain. However, modern browsers are very efficient at this — selectors that never match are filtered out quickly by the rightmost part of the selector (e.g., the browser checks "is this an 
  h1?" before checking if it's inside .uk-section-primary). So the rendering speedup exists but is marginal.                                                                                                       
                                                                                                                                                                                                                 
  The practical impact is mainly in the first category — less CSS to download and parse on every admin page load.

BernhardBaumrock:
If my concerns are the only reason for having those classes, feel free to remove them, as I don't have these concerns any more.

PR #329, opened 2026-05-17
Add `$config->ClassName` array overlay for module config settings   

adrianbj:
## Summary                                                                                                                                                                      

  This PR generalises a pattern several existing modules (notably TracyDebugger) already implement locally: overriding module config values from `site/config.php` via a `$config->ClassName` array. After this change, the override works for **any** configurable module without per-module code.
                                                                                                                                                                                    
  A common use case is reading secrets from `$_ENV` (.env files via phpdotenv etc.) and feeding them into module config without storing them in the database or in version-controlled files:
                                                                                                                                                                                    
  ```php                                                                                                                                                                          
  // site/config.php
  $config->WireMailMailgun = [                                                                                                                                                      
      'apiKey' => $_ENV['MAILGUN_API_KEY'],
      'domain' => $_ENV['MAILGUN_DOMAIN'],                                                                                                                                          
  ];                                                                                                                                                                                
   
  $config->TracyDebugger = [                                                                                                                                                        
      'outputMode' => $_ENV['APP_ENV'] === 'prod' ? 'production' : 'development',                                                                                                 
  ];
  ```                                                                                                                                                                               
   
  ## Behavior                                                                                                                                                                       
                                                                                                                                                                                  
  `Modules::getConfig($class, $property = '')` now overlays values from `$wire->config->ClassName` on top of the stored DB config before returning.

  - Shallow merge of top-level keys (matches the TracyDebugger pattern)                                                                                                             
  - `getConfig('Foo')` returns `array_merge($dbData, (array) $config->Foo)`
  - `getConfig('Foo', 'apiKey')` returns the overlay value when present, else DB value                                                                                              
  - DB rows are **never** mutated by the overlay                                                                                                                                    
  - Internal `configData[$id]` cache continues to store raw DB data; overlay is applied at read time after cache lookup
  - Non-array `$config->ClassName` (string, object, missing) → overlay silently ignored                                                                                             
  - Empty array `$config->ClassName` → no-op                                                                                                                                      
                                                                                                                                                                                    
  Because `ModulesConfigs::setModuleConfigData()` already routes through `getConfig()` when no `$data` argument is provided, module instance properties applied at init time automatically pick up the overlay too — no separate change needed for that path.                                                                                                  
                                                                                                                                                                                    
  ## Admin UI indicator                                                                                                                                                           

  To avoid confusion when an override is active, `getModuleConfigInputfields()` now decorates the module config form:                                                               
   
  - Fields whose names match a key in `$config->ClassName` get a `notes` line appended ("Currently overridden by ...") so admins know edits won't take effect at runtime.           
  - A single `InputfieldMarkup` summary is prepended to the form listing the overridden keys.                                                                                     
  - Override values are **never** rendered (secret-safe). Inputs stay editable so the stored DB value remains useful as a fallback when the override is removed.                    
                                                                                                                                                                                    
  ## Backward compatibility
                                                                                                                                                                                    
  Purely additive. Modules that don't have `$config->ClassName` set behave identically. Modules that already implement their own internal overlay (TracyDebugger) continue to work —
   the merge is idempotent for them.
                                                                                                                                                                                    
  ## Files changed                                                                                                                                                                

  - `wire/core/Modules/ModulesConfigs.php` — `getConfig()` implementation, PHPDoc, and admin UI indicator in `getModuleConfigInputfields()`                                         
  - `wire/core/Modules/Modules.php` — PHPDoc on the wrapper

### DIFF
--- a/wire/core/Modules/Modules.php
+++ b/wire/core/Modules/Modules.php
@@ -1628,27 +1628,34 @@ class Modules extends WireArray implements CliModule {
 
 	/**
 	 * Given a module name, return an associative array of configuration data for it
-	 * 
+	 *
 	 * - Applicable only for modules that support configuration.
 	 * - Configuration data is stored encoded in the database "modules" table "data" field.
-	 * 
+	 * - When `$wire->config->ClassName` is set to an array in `site/config.php`,
+	 *   its top-level keys override the stored DB config on every read. The DB
+	 *   is never mutated by the overlay. (3.0.263+)
+	 *
 	 * ~~~~~~
 	 * // Getting, modifying and saving module config data
 	 * $data = $modules->getConfig('HelloWorld');
 	 * $data['greeting'] = 'Hello World! How are you today?';
 	 * $modules->saveConfig('HelloWorld', $data);
-	 * 
+	 *
 	 * // Getting just one property 'apiKey' from module config data
-	 * $apiKey = $modules->getConfig('HelloWorld', 'apiKey'); 
+	 * $apiKey = $modules->getConfig('HelloWorld', 'apiKey');
+	 *
+	 * // Override config from site/config.php (3.0.263+):
+	 * // $config->HelloWorld = [ 'apiKey' => $_ENV['HELLOWORLD_API_KEY'] ];
 	 * ~~~~~~
 	 *
 	 * #pw-group-configuration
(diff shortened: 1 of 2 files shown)


adrianbj:
@ryancramerdesign - any chance you'd consider prioritizing this one please? I do really think it's important that we get our various access tokes/secrets out of our database (which is often backed up to file and replicated all over the place) and into a single non-version controlled .env file.

Thanks!

PR #335, opened 2026-07-19
WireHttp: consistent redirect handling across send types + followRedirects option (issues#2291)

adrianbj:
Fixes the WireHttp redirect inconsistencies detailed in processwire/processwire-issues#2291: for the same redirecting URL, the reported http code, response headers and body differed by send type, and there was no way to opt out of following.

Adds a `followRedirects` option to `$options` (bool, or int for max hops; default `true`) and normalizes all three transports to one contract:

- **following (default):** reported code and headers are those of the **final** response, on every transport — matching what the curl transport (first in the default `use` order) already reported.
- **`followRedirects => false`:** reported code and headers are those of the **first** response, with its `location` header available, on every transport.

### 1. fopen reported the first hop's status with the final hop's body

`$http_response_header` contains headers from every response in the chain and `setResponseHeader()` parsed the status from element `[0]` — the first hop. A `301 → 200` URL reported 301 while returning the 200 page's body.

**Fix:** the header array is reduced to the final response's block (last `HTTP/` status line onward) before parsing, so code/headers/body agree and match curl. `follow_location` is now also set explicitly from the option, and `ignore_errors` is enabled only in no-follow mode so a redirect response can be read instead of failing `fopen()`.

### 2. curl retained stale headers across hops

`CURLOPT_HEADERFUNCTION` accumulated headers from every response, so a followed `301 → 200` reported code 200 with the first hop's `location` header still present.

**Fix:** the callback resets its array whenever a new `HTTP/` status line begins, leaving only the final response's headers. `CURLOPT_FOLLOWLOCATION` is now set explicitly from the option along with `CURLOPT_MAXREDIRS` (20, matching fopen's default cap, previously unlimited).

### 3. socket permanently stopped following redirects after 5 hops per process

`sendSocket()` used `static $level` as its recursion guard and never reset it, so the count accumulated across all requests in the process — in long-running CLI/queue processes, redirect following silently dies after the 5th followed hop.

**Fix:** the level is carried per-call through `$options` instead of a static.

### 4. socket only followed 301/302

curl and fopen follow 303/307/308 as well. **Fix:** socket now follows the same set.

### 5. `send()` never passed `$options` to `sendSocket()`

A per-call `timeout` option (and now `followRedirects`) was silently lost on the socket path, including on the recursive redirect call. **Fix:** options are passed through both.

### Measured before/after

Local test server, `301 → 200` chain, all of curl/fopen/socket × GET/HEAD × 301/302/307:

Before: curl `200` + stale location header; fopen `301` + final body; socket `200` for 301/302 but `307` unfollowed; socket following dead after 5 lifetime hops.

After (default): all 18 combos report `200`, no stale location header, final body. After (`followRedirects => false`): all 18 combos report the true `301/302/307` with the `location` header set. Socket exhaustion no longer reproduces.

### Backwards compatibility

- Default behavior matches what curl — first in the default `use` order — already did, so typical installs see no change in reported code.
- fopen-only environments change from "first-hop status + final body" to "final status + final body" on redirecting URLs; the old combination was internally inconsistent, so this is treated as a bug fix.
- Code relying on hop-1 headers merged into the final response would change; that merged data was unreliable to begin with.

Practical motivation: link-verification code (e.g. FieldtypeVerifiedURL's `useRedirect`/`fixRedirect` features) needs to see the redirect status itself. With curl following by default, that code never sees a 301 on a typical install today; with this PR it can pass `'followRedirects' => false`, which older WireHttp versions simply ignore.

### DIFF
--- a/wire/core/Tools/WireHttp/WireHttp.php
+++ b/wire/core/Tools/WireHttp/WireHttp.php
@@ -520,6 +520,9 @@ class WireHttp extends Wire {
 	 *    then either call `$http->resetRequest()`, create a new WireHttp instance for each request, or specify this option 
 	 *    as true. 3.0.253+ (default=false)
 	 * - `headers` (array): Add these headers to the request, specify as `[ 'name' => 'value' ]` array. 3.0.253+ (default=[])
+	 * - `followRedirects` (bool|int): Follow redirects? Specify boolean, or int for max redirects to follow. Applies to all
+	 *    send types consistently: when following, reported http code and response headers are those of the final response;
+	 *    when false, those of the first response, with its 'location' header available. (default=true)
 	 * @return bool|string False on failure or string of contents received on success.
 	 *
 	 */
@@ -553,7 +556,7 @@ class WireHttp extends Wire {
 			} else if($use === 'fopen') {
 				$result = $this->sendFopen($url, $method, $options);
 			} else if($use === 'socket') {
-				$result = $this->sendSocket($options['_url'], $method);
+				$result = $this->sendSocket($options['_url'], $method, $options);
 			} else {
 				$error[] = "unrecognized type: $use";
 			}
@@ -585,6 +588,7 @@ class WireHttp extends Wire {
 		$defaults = array(
 			'use' => array('curl', 'fopen', 'socket'),
 			'proxy' => '',
+			'followRedirects' => true,
 			'_url' => $url, // original unmodified URL
 			'allowFopen' => true,
 			'allowCURL' => true,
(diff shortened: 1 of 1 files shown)


ryancramerdesign:
Thanks Adrian — this is excellent work. All five bugs verified in the code, the fixes are correct, and the `followRedirects` option is cleanly wired through all three transports. The before/after matrix in the PR description is a nice touch. Code style matches the rest of the file throughout. Merging this into dev.

-Ryan

ryancramerdesign:
Merged into dev. Thanks Adrian!

PR #339, opened 2026-08-14
FieldtypePage: halve the round trips when deleting a page

adrianbj:
Halves the number of DELETE statements `FieldtypePage::hookPagesDelete()` issues on every page delete.

Reported as processwire/processwire-issues#2304.

### The problem

The hook runs on every `Pages::delete()` and issues two DELETE statements for every FieldtypePage field in the installation — one for inbound references (`data`), one for the page's own references (`pages_id`). The count scales with how many page reference fields the site has, not with the page being deleted.

On a site with **124 FieldtypePage fields**, deleting a single leaf page with no children issued **264 DELETE statements, 248 of them from this hook**.

### The change

One statement per table instead of two:

```php
DELETE FROM `$table` WHERE data=:page_id OR pages_id=:page_id2
```

Both columns are indexed, so MySQL resolves the OR as an index_merge union rather than a scan:

```
EXPLAIN DELETE FROM field_x WHERE data=N OR pages_id=N
type: index_merge   key: data,PRIMARY   Extra: Using union(data,PRIMARY)
```

### Measured

Same site, same page, ProcessWire 3.0.270 / PHP 8.5 / MySQL 8:

| | DELETE statements |
|---|---|
| before | 264 |
| after | **140** |

Exactly 124 fewer — one per page reference field.

### Behaviour unchanged

Verified against a page holding both directions of reference (9 inbound rows across two tables, 5 of its own). Before the delete both sets were present; after it, zero rows referencing that page remained in any FieldtypePage table — identical with and without the patch.

### Caveats

- The win is proportional to the number of page reference fields. A site with a handful sees little.
- It matters most when the database is not on localhost. On AWS RDS, per statement round trip latency dominates and page deletes were measured at ~1.8s each; this halves that.
- On a local socket the merged statement may be marginally slower per execution even though it halves the round trips.
- Tested on MySQL 8 only. I have not verified the index_merge plan on MariaDB.

### Not addressed here

The larger win would be batching across a recursive delete — the hook currently fires once per page, so removing a 338 page subtree issued ~167,000 statements. Collecting ids and issuing one `DELETE ... WHERE data IN (...)` per table per tree would be a much bigger reduction, but it is a more invasive change. Noted in the issue rather than attempted here.

### DIFF
--- a/wire/modules/Fieldtype/FieldtypePage/FieldtypePage.module
+++ b/wire/modules/Fieldtype/FieldtypePage/FieldtypePage.module
@@ -87,15 +87,16 @@ class FieldtypePage extends FieldtypeMulti implements Module, ConfigurableModule
 		$database = $this->wire()->database;
 		foreach($this->wire()->fields->findByType('FieldtypePage') as $field) {
 			$table = $database->escapeTable($field->table);
-			// delete references to this page
-			$query = $database->prepare("DELETE FROM `$table` WHERE data=:page_id");
+			// delete references to this page (data), and references this page is
+			// keeping to other pages (pages_id), in a single statement. Both columns
+			// are indexed, so MySQL resolves this as an index_merge union rather than
+			// a scan. One statement per table instead of two halves the number of
+			// round trips, which dominates when the database is not local.
+			$query = $database->prepare("DELETE FROM `$table` WHERE data=:page_id OR pages_id=:page_id2");
 			$query->bindValue(":page_id", $page_id, \PDO::PARAM_INT);
+			$query->bindValue(":page_id2", $page_id, \PDO::PARAM_INT);
 			$query->execute();
-			// delete references this page is keeping to other pages
-			$query = $database->prepare("DELETE FROM `$table` WHERE pages_id=:page_id");
-			$query->bindValue(":page_id", $page_id, \PDO::PARAM_INT);
-			$query->execute();
-		}	
+		}
 	}
 
 	/**


adrianbj:
Closing the one caveat in the description: the `index_merge` plan is confirmed on **MariaDB**, not just MySQL 8.

Run against a production MariaDB instance (RDS), on a `FieldtypePage` table with the standard schema:

```
EXPLAIN DELETE FROM field_reusable_block WHERE data=12345 OR pages_id=12345;

id  select_type  table                 type          possible_keys  key           key_len  ref   rows  Extra
1   SIMPLE       field_reusable_block  index_merge   PRIMARY,data   data,PRIMARY  4,4      NULL  2     Using union(data,PRIMARY); Using where
```

Same plan as MySQL 8 — `index_merge` with `Using union(data,PRIMARY)`, no table scan. So both engines resolve the merged statement via the two existing indexes, and the change halves the round trips on either.

ryancramerdesign:
Thanks Adrian — this landed on dev as e8766378, applied via cherry-pick so your authorship is preserved on the commit. GitHub won't auto-close the PR since the cherry-pick produces a different SHA, so closing it manually.

Verified the change here before applying it. On a test install with 11 page reference fields, deleting a single page with references in both directions went from 35 DELETE statements to 24 — exactly one fewer per field, as you measured — with identical cleanup.

One note on the query plan, since your caveat mentioned it: my first `EXPLAIN` showed `type=index` rather than `index_merge`, which looked like it might be a regression. That turned out to be an artifact of a nearly empty table, where a full index scan is trivially cheap. On tables with real rows the plan is exactly what you reported:

```
type: index_merge   key: data,PRIMARY   Extra: Using union(data,PRIMARY)   rows: 2
```

Tested on MySQL 8.0.44. MariaDB is still unverified, so your caveat there stands.

Also followed up on the batching you described as "not attempted" — that is now on dev too (58506988), using the existing `Pages::deleteBranchReady` / `Pages::deletedBranch` hooks to collect page IDs and issue one statement per field per branch. Details in processwire/processwire-issues#2304.

-Claude

PR #341, opened 2026-08-26
Prevent browser scroll anchoring from scrolling to bottom of opened repeater items

adrianbj:
Fixes processwire/processwire-issues#2320

## Problem

Opening a repeater / Repeater Matrix item whose ajax-loaded content is large (multi-language fields with language tabs, TinyMCE fields, nested repeaters) leaves the page scrolled so the **bottom** of the item is in view instead of the top.

## Cause

Not ProcessWire's JS — with `window.scrollTo`/`scrollBy`, `Element.scrollIntoView`, jQuery `animate({scrollTop})`/`scrollTop(v)`, focus events and hash changes all instrumented, none fire when the jump happens. It is the browser's native **scroll anchoring**:

1. The item's ajax-loaded content is appended and shown; before LanguageTabs/TinyMCE initialize it is momentarily far taller than its final height (in the traced case the document grew from ~3,700px to ~21,000px, settling at ~10,400px).
2. Scroll anchoring picks an anchor node below the insertion point and silently adjusts `scrollY` by the inserted height to keep it visually stationary, dragging the viewport down.
3. The content then consolidates to its real height and the scroll position clamps — resting at the bottom of the opened item.

## Fix

Opt repeater item lists out of scroll anchoring:

```css
.InputfieldRepeater .Inputfields {
	overflow-anchor: none;
}
```

Added to `InputfieldRepeater.scss` with the same rule appended to the compiled `InputfieldRepeater.css` (only this rule was appended by hand — a full recompile with current dart-sass also rewrites `background:transparent` to `background:rgba(0,0,0,0)` elsewhere, which I avoided to keep the diff minimal).

Verified against the affected install: without the rule the opened item lands at `top: -6506px` relative to the viewport; with it (html-wide or scoped as above) the item opens with its header exactly where it was. `overflow-anchor` is supported by Chrome/Edge and Firefox (the two engines with scroll anchoring); Safari has none and ignores the property harmlessly.


ryancramerdesign:
Thanks Adrian — this landed on dev as 792544a0, applied via cherry-pick so your authorship is preserved on the commit. GitHub won't auto-close the PR since the cherry-pick produces a different SHA, so closing it manually.

The instrumentation in the issue is what made this straightforward to accept. Ruling out `scrollTo`/`scrollBy`, `scrollIntoView`, jQuery `animate({scrollTop})`/`scrollTop()`, focus and hash changes before concluding it was the browser is exactly the work that usually doesn't get done, and it's the difference between a guess and a diagnosis.

Also worth noting you appended the compiled rule rather than regenerating the whole CSS file. That was the right call — I compiled the SCSS here with a current sass and got unrelated differences in the output (`transparent` vs `rgba(0,0,0,0)`, `rgb` vs `hsl`), so a full rebuild would have brought in cosmetic churn that had nothing to do with the fix.

-Claude

PR #343, opened 2026-09-01
AdminThemeUikit: fix login form fields hidden behind oversized masthead

adrianbj:
Fixes processwire/processwire-issues#2331

On the themes/default login screen, the login overrides set `#pw-masthead-mobile { height: auto }` so the logo area can grow, but the parent `#pw-mastheads` keeps the theme's fixed `height: 56px`. With a custom admin logo whose login logo box (`.pw-logo-link`, `aspect-ratio: 4/3`, contents scaled 1.35×) exceeds 56px — e.g. any wide logo such as a 300×66 SVG — the masthead overflows its container and hangs down over the top of the login form. Because the masthead carries `z-index: 100 !important`, it stacks above the Username field (blocking mouse interaction), and on sites whose custom admin CSS gives `.pw-masthead` an opaque background it visually erases the Username label/input and the Password label, leaving only the Password input visible.

This adds `height: auto` to the existing login-screen `#pw-mastheads` override so the logo area pushes the form down instead of overlapping it.

Verified:
- On an affected live site (3.0.271, wide custom logo, opaque `.pw-masthead` background): injecting this one rule restores both fields and labels; `document.elementsFromPoint()` at the Username input's center returns the input itself instead of the `.pw-logo-link` / `#pw-masthead-mobile` stack.
- On a stock 3.0.270 install with the default logo: no visual regression — the form just sits a few pixels lower as the logo box takes its natural in-flow height.


ryancramerdesign:
Merged to dev as e1f70886, cherry-picked so your authorship is preserved. Closing this PR as the commit is now on dev rather than through the merge.

Verified before and after by forcing the masthead child to the 173px height reported in processwire/processwire-issues#2331: it previously overflowed its 56px parent by 117px and covered the Username input, and now the masthead sizes to its content with no overflow. Default logo installs are unaffected apart from the form sitting a few px lower.

Thanks Adrian.

-Claude

PR #344, opened 2026-09-02
Add default value for $yes parameter in $page->if() function, fixing IDE's reporting false error

brodiefarrell:
Added a default value in the class comment for the $yes parameter of the $page->if() function. This stops IDE's from falsely complaining that using `$page->if() `with only a $key parameter is incorrect, when it works.

Fixes issue #2333: [$page->if(...) documentation mismatch for calling with one paramater](https://github.com/processwire/processwire-issues/issues/2333)


brodiefarrell:
Sorry I only just signed my Contributor License Agreement after submitting the PR, not before like the contribution guidelines state. I hope that's okay!

brodiefarrell:
Closing as this has been addressed in [Commit 17e02dd](https://github.com/processwire/processwire/commit/17e02dd20af9a39fe87ea861804a743aed6c8478)

PR #346, opened 2026-09-02
ProcessPageList: load page list children in a single query (+ findMin UTF-8 name fix)

adrianbj:
Two commits that land together: a one-query loading strategy for page list children, and the `PagesLoader::findMin()` fix it depends on.

### 1. Fix: findMin() leaves default-language UTF-8 page names punycode-encoded

With `$config->pageNameCharset = 'UTF8'`, page names are stored punycode-encoded (`xn--`) and decoded on load. `findMin()` (any find with `field=`/`join=` in the selector, 3.0.172+) already decodes the language name columns (`name1234`) via `setName()`, but routes the default `name` column through `setForced()`, which skips the decode — so pages loaded this way report punycode paths/urls (`/xn----ctbjbl5bhngbc/` instead of the UTF-8 name), differing from the same page loaded via `getById()`. This is a pre-existing bug independent of the second commit, fixed by routing punycode default names through `setName()` like the language columns.

### 2. ProcessPageList: one query for children instead of one per template

`PagesLoader::getById()` groups page loads by template (to apply each template's autojoin fields), so a page list branch whose children span many templates costs close to one query per listed page — e.g. a home branch with 50 children across 37 templates ran 37 load queries, for superusers and editors alike. `getSelector()` now appends `join=title`, which routes child loading through `findMin()`: one query regardless of template count, with `title` joined for page labels. Any additional label fields lazy-load exactly as before, and the hookable `getSelector()` lets sites join more fields or remove the join.

`getSelector()` also makes the default sort explicit (the same sort `children()` would apply) and appends `sort=id` as a secondary sort, making ordering and pagination deterministic when the primary sort has duplicate values — previously, dozens of trashed pages sharing one `modified` timestamp could shift between pagination windows depending on query plan.

### Measurements

Real site, 16.5k pages, page tree branch requests (queries via `$database->queryLog()` around `ProcessPageList::execute()`):

| Request | Before | After |
|---|---|---|
| home branch (58 children, 37 templates) as superuser | 42 queries | **7** |
| same as editor (with #345's batched counts) | 95 queries | **13** |
| branch of 49 pages that all have children, as editor | 156 queries | **10** |
| initial load with 4 remembered-open branches | 289 queries | 240 |

### Verification

- All 592 branches of the site rendered byte-identical vs the previous strategy, as superuser and as a non-superuser editor, in default and a second language — the only difference anywhere is the intentional deterministic ordering among equal sort values (verified stable across repeated runs).
- Pagination boundaries (partial last pages, deep offsets), the `limit=0` selector used by ajax actions, and `openPageData` pre-rendering: identical output.
- Multi-language install (FieldtypePageTitleLanguage + LanguageSupportPageNames + UTF8 page names): language titles, language-specific names, and punycode names in both default and non-default language columns render identically to the previous strategy (the punycode default-name case is what surfaced the findMin bug in commit 1).
- Child count bubbles and pagination totals unaffected (`numChildren` rides along in the finder's verbose rows).

Independent of but complementary to #345 (batched child counts): no overlapping files; together they take the editor home branch from 95 queries to 13.


adrianbj:
Added a guard to the join in a commit amend: `join=title` is now skipped when the branch has more than 1000 children. Benchmarking found that the single findMin query scans/sorts all siblings with the joined data, which on a 3113-child single-template branch cost ~12ms more than the two queries it replaced (~5µs per sibling row, while each saved per-template load query is worth ~0.2ms locally and a full round trip in production). Branches that large are effectively never multi-template, so they keep the previous strategy. With the guard, that branch benchmarks at parity with dev, and multi-template branches keep the full win (home branch as superuser: 42 → 7 queries; interleaved medians 186ms → 159ms as editor with #345, large-branch initial loads ~400ms → ~330ms).

PR #353, opened 2026-09-09
Build repeater item inputfields from unformatted values

adrianbj:
Fixes processwire/processwire-issues#2343 (and the render-side cause of processwire/processwire-issues#2249)

### Problem

`FieldtypeRepeater::formatValue()` enables output formatting on the repeater item Page objects themselves (`$p->of(true)`), and those objects are shared with the unformatted value. So whenever anything gets a formatted value of a repeater field while the page editor renders — typically a hook or module building an SEO preview, sitemap entry, etc. — the item pages are left output-formatted, and `InputfieldRepeater` then builds every input in the item from FORMATTED values:

- Hanna Code tags are replaced with their rendered output (#2343)
- entity-encoded text is re-encoded
- multi-language values collapse to a plain string in the current language, so non-default language inputs render empty (the prefill failure reported in #2249)

The database is still correct at this point, but saving the page writes the formatted values back, silently destroying the raw values.

This also explains the reporter's observation that only an *old* repeater was affected while a new repeater with the same field was fine: it isn't legacy field config — only the old field is referenced by whatever module does a formatted get during the edit render, so nothing flips the new field's items.

### Fix

`InputfieldRepeater::getRepeaterItemInputfields()` now turns off output formatting on the item while building its inputfields and restores it afterwards — the same guard `formToPage()` already applies on the input-processing side. All item inputfield building funnels through this method, so it covers regular repeaters, FieldsetPage (single-item mode) and ajax-loaded items. RepeaterMatrix inherits it too, provided it doesn't override the method.

### Tests

Added a regression test to `FieldtypeRepeater.test.php` (using the existing `TextformatterEntities` sub-field as a stand-in for Hanna Code): it gets a formatted value of the repeater field, then renders the field's inputfield via the public `getInputfield()->render()` path and asserts the item input contains the raw value (`value="Ampersand &amp; Test"`) rather than the formatted one (`value="Ampersand &amp;amp; Test"`).

- fails against the previous code with "Item inputfield was built from FORMATTED value"
- passes with the fix
- a full `test all` run is identical before and after the change


ryancramerdesign:
Thanks @adrianbj — your diagnosis and patch were incorporated in commit 8faf7c4b, with you credited as co-author. I added exception-safe output-formatting restoration and made the regression fixture explicitly ensure and restore TextformatterEntities while also checking restoration of the item state. Closing this PR as incorporated.

PR #360, opened 2026-09-23
Add WireDatabaseDialect::upsert() and use it for core insert-or-update queries

adrianbj:
Requested in processwire/processwire-requests#609 ([comment](https://github.com/processwire/processwire-requests/issues/609#issuecomment-5801302021)): the first of the dialect SQL helpers, chosen because it is the one case where translation cannot do the job. `INSERT ... ON DUPLICATE KEY UPDATE` needs no key information on MySQL, but `ON CONFLICT (...) DO UPDATE SET` needs the unique key the insert conflicts on, and a translator cannot infer that from the statement (the SQLite translation currently emits `ON CONFLICT` with no target for exactly this reason).

## What changed

**`WireDatabaseDialect::upsert($table, $columns, $update, $options)`** builds the insert-or-update statement in the dialect's own syntax.

- `$columns`: list of names (bound as `:name`) and/or `name => expression`.
- `$update`: list of names (set to the inserted value; must be among `$columns`) and/or `name => expression` (`'NOW()'`, `'qty+1'`, `':bind'`).
- `$options['conflict']`: the key columns the insert conflicts on. `$options['rows']`: multi-row literal inserts.
- Base implementation (MySQL): `ON DUPLICATE KEY UPDATE col=VALUES(col)`. SQLite override: `ON CONFLICT (cols) DO UPDATE SET col=excluded.col`, target omitted when the caller doesn't give one.
- Throws `WireDatabaseException` for nothing to insert, nothing to update, an update column that isn't inserted, or a row of the wrong length.

**`nativeSql()` / `isNativeSql()`** on the dialect, and a bypass in `WireDatabasePDO::translatedQuery()`. Helper output is registered as native and executed without translation. This is needed, not just an optimization: the SQLite translator's `insert()` treats any top-level `SET` as the `INSERT ... SET` form, so a native `ON CONFLICT ... DO UPDATE SET` clause fails with `near "(": syntax error` if it reaches the translator (the new test reproduced this before the bypass existed). Non-translating dialects treat all SQL as native, so MySQL is unaffected. I'm not attached to the registry mechanism if you'd prefer something else (a `prepare()` option, a marker, or a query object flag).

**15 core call sites migrated** (all of the runtime `ON DUPLICATE KEY UPDATE` statements): `Fieldtype::savePageField()` (both branches), `Modules` info cache, `PagesAccess` (single and multi-row), `PagesEditor::sortRebuild()`, `PagesSortfields`, `Users::setAdminThemeByRole()`, `WireCacheDatabase`, `WireDataDB`, `FieldtypeFieldsetPage`, `PagePaths`, `PagesVersions` (two), `SessionHandlerDB`. The two remaining occurrences are in `WireDatabaseBackup`, which writes MySQL dump text rather than executing statements, so they are left as they are.

**Conflict target policy.** Call sites pass the key when the table has one candidate key for the insert (`pages_id` for field tables, `name` for `caches`, `class` for `modules`, `(source_id, name)` for WireDataDB, `(pages_id, field_id, version)` for version field data, `id` for sessions and for `pages` in `sortRebuild()`). Where a table has more than one unique key that the insert could hit (`pages_paths`: PK plus unique `path`; `pages_versions`: PK plus unique `(name, pages_id)`), no target is passed, with a comment saying why: MySQL's `ON DUPLICATE KEY` covers every key while a conflict target covers one, so passing one would change SQLite behaviour in those edge cases. That decision will need revisiting for a database that requires a target.

**MySQL output.** Semantically identical at every site; formatting is normalized (all identifiers backticked, `VALUES (` spacing), which the tests pin. Three sites drop redundant placeholders as a result: `Users` and `PagesSortfields` no longer bind separate `:update_*` values for what is the inserted value anyway, and `SessionHandlerDB` no longer reuses the same named placeholders in the update clause (which relied on emulated prepares).

**Docs.** `WireDatabase/API.md` gets an "SQL helpers" section for `upsert()` and the native SQL rule; `Fieldtype/API.md` updated where it named the MySQL syntax.

## Tests

New `wire/core/WireDatabase/WireDatabaseDialect.test.php` (38 checks): pinned MySQL output for each statement shape used by the migrated sites, pinned SQLite output with and without a conflict target, the native/non-native distinction, and live insert-then-update, expression-update, bound-update and multi-row checks on the site's own database. The MySQL and SQLite dialect checks construct the dialect directly, so they run on either database.

| Run | Before | After |
|---|---|---|
| WireTests `all`, MySQL 8 (existing site) | 102 passed, 4 failed | 103 passed, 4 failed (same four: LanguageSupport, MarkupPagerNav, ProcessWire, TemplateFile, all environmental on that site) |
| WireTests `all`, SQLite (fresh blank-profile CLI install from this branch) | n/a | 104 passed, 1 failed (WireHttp expects HTTP 200 from a site with no web server) |
| `WireDatabaseDialect` test | n/a | 38/38 on MySQL, 38/38 on SQLite |

Also exercised by script on both databases: page saves (pages_access), sortfield set/changed/cleared (pages_sortfields insert, update, delete), `modules->refresh()`, cache save-twice, `sortRebuild()`.

## Pre-existing SQLite issue found while testing

`PagesEditor::sortRebuild()` fails on SQLite with `NOT NULL constraint failed: pages.name`, both before and after this change (verified with the original `INSERT INTO pages (id,sort) ... ON DUPLICATE KEY UPDATE` text). SQLite evaluates NOT NULL constraints before upsert conflict resolution, so an insert that omits `name` fails even though every row exists and would only be updated. MySQL accepts it because strict mode is off. It's the "upsert as bulk update" pattern; a portable form is an `UPDATE ... SET sort=CASE id ... END` or per-row updates. Not addressed here to keep this PR to the helper; happy to follow up separately.


ryancramerdesign:
Thanks Adrian — I pulled the branch and ran it. WireTests on my end: 107/108 on SQLite (only an environmental failure) and 105/105 on MySQL, with your dialect test at 38/38 on both. A few things I checked rather than took on faith:

- Every `conflict` target you pass matches an actual unique key. I verified each against a live SQLite schema, including `modules`, where the target is the unique `class` index rather than the primary key — that's the case where a wrong target throws at runtime.
- `SessionHandlerDB`'s `['ts' => 'NOW()']` resolves on SQLite, since helper output skips translation and `NOW()` is one of the functions we register on the connection.
- The migrated call sites keep their semantics. Dropping `pages_id=:update_pages_id` in `Users::setAdminThemeByRole()` is right, and no longer binding the same placeholder twice in SessionHandlerDB is an improvement on what was there.

**Your call to leave the conflict target off for `pages_paths` and `pages_versions` is the right one.** MySQL's `ON DUPLICATE KEY` covers every unique key and SQLite's bare `ON CONFLICT` does the same, so naming one key would have narrowed the behaviour. Good that you surfaced it instead of picking one.

You're right about the counts, too — my 22 was occurrences including documentation and comments.

**One change I'd like before merging: drop `nativeSql()` / `isNativeSql()`.**

You correctly identified that the translator mangles a native `ON CONFLICT` statement, but that was a bug in the translator rather than something to route around: `insert()` treated any top-level `SET` as the `INSERT ... SET` form, including the one inside `DO UPDATE SET`. That's fixed on dev in 74a88d4d, with a test — a native upsert statement now passes through translation unchanged.

With that fixed, the registry saves about 0.1µs on an already-cached translation, and in exchange the SQL has to be used byte-for-byte: appending a single space silently sends it through the translator instead, which your own test documents. I'd rather not have that edge in the API for that much gain. If you rebase onto dev and remove the mechanism and its `translatedQuery()` branch, I think this is ready.

Smaller note: the `rows` option interpolates values as raw SQL. The call sites all cast to int and the docblock does say "SQL value expressions", but I'd like that warning to be louder, or the option restricted to scalars.

**On `sortRebuild()`:** confirmed, and it's pre-existing — it fails the same way with the original statement, with yours, and through the API. Yes please to the follow-up, and the `UPDATE ... SET sort=CASE id ... END` approach sounds right. Worth a test with it, since nothing in WireTests currently touches that method, which is why we hadn't seen it.

On Postgres and `ONLY_FULL_GROUP_BY`: you're right about functional dependency, and that does cover `GROUP BY pages.id` with columns from the pages table. The joined-table sort columns are the case I'd want to see proven.

-Claude

adrianbj:
Thanks Ryan, both changes are in and the branch is rebased onto dev (74a88d4d).

- `nativeSql()` / `isNativeSql()` are gone, along with the branch in `translatedQuery()` and the API.md paragraph. You were right that it was a translator bug rather than something to route around; with your fix the SQLite dialect's `upsert()` output goes through translation and comes out unchanged, and the dialect test's live section confirms it end to end on SQLite.
- `rows` now accepts scalars only. Ints and floats are inserted as given, strings are quoted through `$database->quote()`, `true`/`false` become 1/0, `null` becomes NULL, and anything else throws. Three new checks pin that down, including one showing that a string `'NOW()'` is inserted as the literal string, not the function. The docblock and API.md say the same.

Dialect test is now 35/35 (the six registry checks removed, three `rows` checks added). Full WireTests unchanged from before: MySQL 103 passed, 4 failed of 107, SQLite 104 passed, 1 failed of 105, both with only the environmental failures you'd expect.

The `sortRebuild()` follow-up is #361 (rebased on top of this), with the `CASE id` update and a `testSortRebuild()` in `Pages.test.php` that was red on SQLite before the fix, so that method is covered now.

On the Postgres side, the spike results are on the requests issue; the joined-table sort columns were exactly the case that fails and `any_value()` is the fix that keeps MySQL's semantics.

ryancramerdesign:
Thanks Adrian — that was quick. I've had a look at both this and #361 and they check out on my end: suites green on SQLite and MySQL, and your `testSortRebuild()` does fail against unfixed dev with the NOT NULL error, so it's a real regression test rather than a passing formality.

We're out of the office today, so I'll work with both PRs properly tomorrow rather than merging in a hurry.

One thing worth knowing in the meantime, since it may matter for the Postgres dialect: the two changes here had to land together. On SQLite `$database->quote()` deliberately produces MySQL-style escaping, because the translator converts it — so a string like `it's \ tricky` quoted into SQL that skips translation is a syntax error. Removing the bypass is what makes the new `rows` quoting safe. Worth remembering for a dialect that doesn't translate at all, where helper SQL will need that dialect's own quoting.

-Claude

